PR #4, opened 2016-06-21
For Ubuntu Xenial install Trolltech.conf to ~/.config to avoid conflict

unman:
Also writes Trolltech.conf to skel in case of new user.


marmarek:
Re travis failed build: you forgot to update rpm spec file.

marmarek:
@unman 
How about existing users? Moving file from `/etc/xdg/Trollech.conf` to `/etc/skel` will remove it from existing installations, but not create in users home, in none of existing AppVMs based on such template.
Maybe there is some other solution than `Trolltech.conf` for the original issue? Or maybe Ubuntu default file is good enough and qubes package may not ship it there at all (but still have it in standard debian package)?

PR #5, opened 2016-07-01
Support for pulseaudio 9.0

jenrik:
This adds support for pulseaudio 9.0 and fixes archlinux PKGBUILD failing due to "incompatible" pulseaudio version


jenrik:
Please note that pulseaudio/pulsecore 9.0 brings in some files from Chromium OS that are licensed under a BSD-style license. The affected files are:

- pulse/pulsecore-9.0/filter/biquad.h
- pulse/pulsecore-9.0/filter/crossover.h

marmarek:
> some files from Chromium OS that are licensed under a BSD-style license.

Shouldn't be a problem in this direction.

I wonder if Lennart ever will expose modules API externally (not necessary stable API), to avoid this all nonsense of copying pulseaudio includes...

marmarek:
From where did you get those (all) headers? It looks to have few more commits after 9.0 release - for example this one: https://cgit.freedesktop.org/pulseaudio/pulseaudio/commit/?id=694662d9363d1f61a1021e45fab80ca1e81f3ab1
Was that intentional?

jenrik:
Headers from my previous commit originated from the latest commit to pulseaudio master branch, I accidentally forgot to checkout the v9.0 tag before copy the headers over. I have included the correct version of the headers in commit 86ca8debac3fdd5fe6bdac3806bbdc94a9303315

PR #7, opened 2016-11-07
Exclude Trolltech.conf in Xenial build

unman:
After trying various alternatives, it seems to me that the cleanest solution here is just to do nothing for Xenial, and to use the Ubuntu Trolltech.conf without override. If there are display issues, (and I haven't seen any in testing), it's just a standard Trolltech problem, easily resolvable by user.


marmarek:
@unman is this reopening a try to restart travis build? I've just triggered it manually.

unman:
@marmarek Yes: is there a smarter way, other than asking you to rebuild? I think there's an open issue about handling network issues gracefully, which could help with some of these.

marmarek:
> @marmarek Yes: is there a smarter way, other than asking you to rebuild?

I don't think so...

> I think there's an open issue about handling network issues gracefully, which could help with some of these.

Yes, https://github.com/QubesOS/qubes-issues/issues/1424, there is also link to a script, but I never had a time to integrate it.

PR #8, opened 2016-11-30
Provide XDG_DATA_DIRS in Xenial. Also update deb package to 3.0 quilt)

unman:
This should be final part of xenial build, fixing QubesOS/qubes-issues#2430

Package format has been changed to use quilt. The debian changelog will need to be updated as part of final package build.

### DIFF
--- a/Makefile.builder
+++ b/Makefile.builder
@@ -2,4 +2,18 @@ ifeq ($(PACKAGE_SET),vm)
 RPM_SPEC_FILES := rpm_spec/gui-vm.spec
 ARCH_BUILD_DIRS := archlinux
 DEBIAN_BUILD_DIRS := debian
+ifneq ($(filter $(DISTRIBUTION), debian qubuntu),)
+SOURCE_COPY_IN := source-debian-quilt-copy-in
+
+source-debian-quilt-copy-in: VERSION = $(shell cat $(ORIG_SRC)/version)
+source-debian-quilt-copy-in: ORIG_FILE = "$(CHROOT_DIR)/$(DIST_SRC)/../qubes-gui-agent_$(VERSION).orig.tar.gz"
+ifneq (,$(findstring $(DIST),xenial))
+source-debian-quilt-copy-in:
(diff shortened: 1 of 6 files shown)


marmarek:
Travis:
```
(cd xf86-input-mfndev; if [ -e Makefile ] ; then make distclean; fi; ./bootstrap --clean || echo )

   dh_clean

 dpkg-source -b gui-agent-linux

dpkg-source: error: can't build with source format '3.0 (quilt)': no upstream tarball found at ../qubes-gui-agent_3.2.9+deb9u1.orig.tar.{bz2,gz,lzma,xz}
```
Take a look at other packages in quilt format. And when adding tar call, see https://reproducible-builds.org/docs/archives/

unman:
That error arises because of the versioning - Debian expects [epoch:]upstream_version[-debian_revision]. The upstream tarball is  qubes-gui-agent_3.2.9.tar.xz, but without the deb revision dpkg-build thinks the revision is  qubes-gui-agent_3.2.9+deb9u1, and looks for correspondingly named file. If you add a versioning entry in the debian changelog for qubes-gui-agent_3.2.9-1, the problem goes away. (I mentioned this in my PR)

On the use of tar I just filched this from other packages. Perhaps this should be addressed by raising a separate issue and reviewing for consistency across the board? (Or as part of QubesOS/qubes-issues#1244 which has been hanging for some time?)

marmarek:
Ah, so after bumping package version (in new format) build issue should be gone, right?

As for tar - yes, if no need to change it right now, it's ok to do this separately.

unman:
> Ah, so after bumping package version (in new format) build issue should be gone, right?

Not quite: the debian/changelog file should have an entry at top to match `version` - 
e.g for version 3.2.9
```
qubes-gui-agent (3.2.9-1) jessie; urgency=medium

    * New upstream release
```
This could be generated by `dch -v version-1`

marmarek:
Yes, this is what I meant:
https://github.com/QubesOS/qubes-builder/blob/master/doc/ReleaseManagerWorkflow.md#incrementing-version
(in case of changing version format, I may need to manually fix what
script will produce)

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab
A: Because it messes up the order in which people normally read text.
Q: Why is top-posting such a bad thing?

unman:
Just so

PR #12, opened 2017-03-23
xf86-input-mfndev: don't access windows in input thread

HW42:
Newer Xsevers have a separate thread for input processing. Therefore we should no longer directly access window objects because they can change while we are running. To fix this dump the window memory MFNs in the main thread via QueueWorkProc().

Most likely fixes QubesOS/qubes-issues#2617.

Since the code was a bit messy formated (mix between tabs/spaces and ts=4/ts=8, trailing whitespace, old out-commented code), I first cleaned it up. If you don't like it I can rebase it.


marmarek:
Thanks!

PR #18, opened 2017-09-29
Add pulseaudio-11.1 headers symlink for release3.2 branch

na--:
Cherry-picked commit 1e4c311842fea84919d09950d602ad529b6f5caf - this is needed for building the archlinux template for Qubes 3.2.

### DIFF
--- a/pulse/pulsecore-11.1
+++ b/pulse/pulsecore-11.1
@@ -1,0 +1,1 @@
+pulsecore-11.0


na--:
Not sure why the code-signing check fails - the commit is signed. Or is it the case that it checks if my key is in a list or signed by a particular other key as a valid developer key?

marmarek:
It fetches the key from keyservers and it fails sometimes.

marmarek:
Anyway I prefer `cherry-pick -x` for stable branches (this include original commit id). Will do in a moment.

marmarek:
Done.

na--:
Thanks, will keep that in mind for future PRs!

PR #20, opened 2017-10-23
Archlinux r4 template

ptitdoc:


archlinux/PKGBUILD-pulseaudio.install
```diff
@@ -0,0 +1,21 @@
...
+  ldconfig
+}
+
+post_upgrade() {
+  post_install
+}
+
+## arg 1:  the old package version
+pre_remove() {
+  systemctl disable qubes-gui-agent.service
```
marmarek: This belong to main package, not pulseaudio.

archlinux/PKGBUILD
```diff
@@ -4,29 +4,27 @@
 # then please put 'unknown'.
 
 # Maintainer: Olivier Medoc <o_medoc@yahoo.fr>
-pkgname=qubes-vm-gui
+pkgname=(qubes-vm-gui qubes-vm-pulseaudio)
```
marmarek: How upgrades are handled in such case?

archlinux/PKGBUILD-pulseaudio.install
ptitdoc: Thanks I missed that.

archlinux/PKGBUILD
ptitdoc: You are right, upgrades are not handled in this case and lot of users will complain.

I need to think if it should be kept optional, create a dependency, or use a meta package as you suggested.

I try to not use dependencies because of pulseaudio / xorg upgrade issues. I would like to keep the possibility to uninstal only the pulseaudio component without breaking the package manager.

archlinux/PKGBUILD
```diff
@@ -75,6 +79,18 @@ install -D $srcdir/PKGBUILD-z-qubes-session.sh $pkgdir/etc/X11/xinit/xinitrc.d/z
 
 }
 
+package_qubes-vm-pulseaudio() {
+
+pkgdesc="Pulseaudio support for Qubes VM"
+depends=( 'alsa-lib' 'alsa-utils' 'pulseaudio-alsa'
+		'pulseaudio>=11.1' 'pulseaudio<11.2'
```
marmarek: I wonder if this could be calculated automatically at package build time (see spec file for rpm). But it isn't blocking this PR.

archlinux/PKGBUILD
marmarek: Ok, since documentation now (will) include mention of extra package for sound support, IMO it is ok to simply split the package here.

archlinux/PKGBUILD
ptitdoc: It can be but I never found how to compute next minor versions using bash without leaving a bug for rare (and consequently occuring) cases.

PR #24, opened 2017-11-14
Fix Ubuntu template builds; Apply XDG changes to trusty.

noptys:
4.0 template builds use `<package>.install` files with dh_install.  The
differences between Debian and Ubuntu packages also need to be represented
in these files.

The XDG changes for Xenial are also beneficial in Trusty.  Applying to
both flavors allowed for a more standard Makefile.builder layout.

### DIFF
--- a/Makefile.builder
+++ b/Makefile.builder
@@ -1,11 +1,22 @@
 ifeq ($(PACKAGE_SET),vm)
-RPM_SPEC_FILES := rpm_spec/gui-agent.spec
-ARCH_BUILD_DIRS := archlinux
-DEBIAN_BUILD_DIRS := debian
+  RPM_SPEC_FILES := rpm_spec/gui-agent.spec
 
-ifneq (,$(findstring $(DIST),xenial))
-SOURCE_COPY_IN := source-debian-quilt-copy-in
+  ifneq ($(filter $(DISTRIBUTION), debian qubuntu),)
+    DEBIAN_BUILD_DIRS := debian
+    SOURCE_COPY_IN := source-debian-quilt-copy-in
+  endif
+ 
+  ARCH_BUILD_DIRS := archlinux
+endif
+
+source-debian-quilt-copy-in: VERSION = $(shell cat $(ORIG_SRC)/version)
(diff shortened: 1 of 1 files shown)


Makefile.builder
```diff
@@ -1,11 +1,22 @@
...
+  ARCH_BUILD_DIRS := archlinux
+endif
+
+source-debian-quilt-copy-in: VERSION = $(shell cat $(ORIG_SRC)/version)
+source-debian-quilt-copy-in: ORIG_FILE = "$(CHROOT_DIR)/$(DIST_SRC)/../qubes-gui-agent_$(VERSION).orig.tar.gz"
 source-debian-quilt-copy-in:
+	if [ $(DISTRIBUTION) == qubuntu ] ; then \
+                sed -i /Trolltech.conf/d $(CHROOT_DIR)/$(DIST_SRC)/debian/qubes-gui-agent.install ;\
+	fi
 	-$(shell $(ORIG_SRC)/debian-quilt $(ORIG_SRC)/series-debian-vm.conf $(CHROOT_DIR)/$(DIST_SRC)/debian/patches)
```
marmarek: If this line is to be applied to every debian-based build, please commit `debian/patches/series` directly (and move patches), instead of generating it at build time. And do not generate orig.tar.gz manually (builder-debian will take care of it).
On the other hand, now you've applied it also to non-Ubuntu build (contrary to the commit message).

unman: Also, it's not necessary to make the change in handling Trolltech.conf in every Ubuntu version.

noptys: @marmarek Ack.  That should not be applied to all debian-based builds--it is a mistake.  I will fix.  

Generating orig.tar.gz manually was following the convention of the `Makefile.builder` logic that was in `core-agent-linux` because it seemed to be the preferred method.  Should it not be there either?

noptys: @unman The Trolltech.conf` file doesn't isn't installed for any Ubuntu build currently:

https://github.com/QubesOS/qubes-gui-agent-linux/blob/4cf2d1181be8ce3207c316124db4a2f92850955a/Makefile#L148-L151

Because it is included in `qubes-gui-agent.install`, the build fails for Ubuntu flavors.

noptys:
Please find suggested changes in https://github.com/QubesOS/qubes-gui-agent-linux/pull/25

PR #25, opened 2017-11-15
Fix Ubuntu template builds; apply XDG changes for all Ubuntu templates.

noptys:
4.0 template builds use `<package>.install` files with dh_install.
The differences between Debian and Ubuntu packages also need to be
represented in these files.

The XDG changes for Xenial are also beneficial in Trusty.


noptys:
I verified that the XDG patch applied for Trusty and Xenial, but not Jessie debs.  Also that the `etc/xdg/Trolltech.conf` config exists in Jessie, but not Trusty or Xenial final debs.

PR #26, opened 2017-11-18
Add support for XFCE desktop in appvm

fepitre:


zander:
This unfortunately breaks running anything not xfce. Specifically the `50-xfce-desktop.sh` is guilty of breaking my KDE desktop look and feel.

Is there a way to get these variables from dom0 instead?

It would make more sense to detect what the actual user is running instead of deciding at packaging time.

fepitre:
You have installed the package then switched to KDE?

zander:
The file mentioned in above comment is part of the qubes-vm-gui package (4.0.7-8) now, which comes default with my template (I build the template myself). Can't remove it without breaking Qubes.

The gui package does all x11 handling, so I expect it should stay desktop-environment agnostic.

fepitre:
The file you cited, is provided by qubes-gui-agent-linux-xfce which is installed only when you build a template with XFCE flavor: https://github.com/QubesOS/qubes-builder-fedora/blob/master/template_scripts/packages_qubes_xfce.list https://github.com/QubesOS/qubes-builder-fedora/blob/00420a3b5f11df956f61b82a93bc7ea4f3d906e6/template_scripts/04_install_qubes.sh#L27

I don't see where it could be installed if you don't select this package specifically.

zander:
I have no idea what 'flavors' are. Here is my builder.conf, I followed the most simple building instructions.

[builder.conf.zip](https://github.com/QubesOS/qubes-gui-agent-linux/files/1603169/builder.conf.zip)

fepitre:
Flavor means, for example, minimal, full loaded, gnome, kde etc. Something qualifying the template.
I see that your template is Archlinux. I think the problem is related to the fact that the file falls in the general package whereas it should have its own package like for Debian and Fedora. Would you like to do it? I can do it but not until this week end (lot of work).

zander:
On Thursday, 4 January 2018 11:30:51 GMT Frédéric Pierret wrote:
> Flavor means, for example, minimal, full loaded, gnome, kde etc. Something
> qualifying the template. I see that your template is Archlinux. I think
> the problem is related to the fact that the file falls in the general
> package whereas it should have its own package like for Debian and
> Fedora. Would you like to do it? I can do it but not until this week end
> (lot of work).

Flovors sounds like a good idea,
having this on all distro templates definitely makes sense, and I'm a bit 
worried its missing from the archlinux one.
Could it be done in a distro-agnostic manner?

Either way, as I don't know the building system very well, it would be nice 
if you could help out getting flavors supported in archlinux.

fepitre:
Ok. I will do the packaging for XFCE this week end. Currently there is only gnome which is what you already have and XFCE. In fact I already did it for KDE and LXQT but I oriented my self to release only XFCE. I was hoping some help from other people. I will see what I can do.

PR #27, opened 2017-11-30
CLIPBOARD_DATA: send window id, and ignore window id in incoming pastes

cfcs:
This PR is linked to a matching PR to `qubes-gui-daemon`, see https://github.com/QubesOS/qubes-gui-daemon/pull/15/ for background.

It makes `send_clipboard_data` send the window id in the `window` field and the length of copied data in the `untrusted_len` field rather than sending the length in both.

It also touches on `handle_clipboard_data` to make it use the `untrusted_len` field for determining how much data to copy instead of using the `window` field. I also changed `len` in that function signature to take an `unsigned int` instead of a `signed int`, since that seemed more appropriate (please bear with me if that was incorrect).

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -709,15 +709,15 @@ void process_xevent_configure(Ghandles * g, XID window,
     send_pixmap_mfns(g, window);
 }
 
-void send_clipboard_data(libvchan_t *vchan, char *data, int len)
+void send_clipboard_data(libvchan_t *vchan, XID window, char *data, uint32_t len)
 {
     struct msg_hdr hdr;
     hdr.type = MSG_CLIPBOARD_DATA;
+    hdr.window = window;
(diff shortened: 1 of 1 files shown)


marmarek:
Since this is protocol change, it can't be merged for Qubes 4.0 this late. But generally it looks ok - lets have it scheduled for Qubes 4.1.

marmarek:
Also, for easier upgrade path, better create new message (`MSG_CLIPBOARD_DATA2`?) instead of changing existing one. And have old one either handled the old way, or ignored.

cfcs:
I consider this a bugfix, but I guess 4.1 is better than nothing. :)
I'm on 3.2 and hope to stay there for at least another year while 4.0 stabilizes and the Fedora in dom0 gets EOL'ed so I don't have to constantly download new updates, so I don't have time to wait for 4.1, I'll find another solution.

PR #32, opened 2018-01-06
desktop flavor: specific XFCE make target

fepitre:
I put the XFCE script in a separate make target. Notably because of Archlinux. If I understand correctly the packaging of Archlinux, each package is made with specific make target. It will allow to do the packaging for the XFCE script with simply make install-xfce. For Fedora and Debian, we could think in the future for a make target like install-flavors. What do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,10 +96,12 @@ install-rh-agent: appvm install-common
 		$(DESTDIR)/etc/X11/xinit/xinitrc.d/20qt-x11-no-mitshm.sh
 	install -D appvm-scripts/etc/X11/xinit/xinitrc.d/20qt-gnome-desktop-session-id.sh \
 		$(DESTDIR)/etc/X11/xinit/xinitrc.d/20qt-gnome-desktop-session-id.sh
-	install -D appvm-scripts/etc/X11/xinit/xinitrc.d/50-xfce-desktop.sh \
-                $(DESTDIR)/etc/X11/xinit/xinitrc.d/50-xfce-desktop.sh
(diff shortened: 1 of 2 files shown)


marmarek:
Targets like install-flavors does not exactly match what is needed here - packages are not always split the same way as template flavors. But generally having separate make targets install-something for each subpackage seems like a good idea.

As for this PR, I understand it is WIP, preparation for Archlinux and/or Debian packaging for xfce flavor, right?

fepitre:
Yes this PR fix the problem for archlinux. I will include further commis for the XFCE templates. This week if I have time.

fepitre:
Also, for that I need to do the packaging for core-agent-linux (for Archlinux) and in the builders Debian and Archlinux. Maybe you could accept the fix to prevent Archlinux to be broken? Need time to do all of this.

marmarek:
This commit does not touch archlinux packaging, `install-xfce` isn't used there yet, so I don't see what is broken there right now. Have I misunderstood something?

fepitre:
If I understand, the problem related in https://github.com/QubesOS/qubes-gui-agent-linux/pull/26 by @zander is due to https://github.com/QubesOS/qubes-gui-agent-linux/blob/a3a6d774773510bf145f70a6e345cd0ab4183cb7/archlinux/PKGBUILD#L67 which includes every files processed in make install-rh-agent into the gui-agent-package of Archlinux.

marmarek:
Ah, I see.

fepitre:
So until I found some time to do XFCE flavor for Archlinux and Debian it would solve the problem.

PR #33, opened 2018-02-01
Using the default_user available in qubesdb to start X

yilmi:
Using the default_user available in qubesdb to start X to fix issue https://github.com/QubesOS/qubes-issues/issues/2372

### DIFF
--- a/appvm-scripts/usrbin/qubes-run-xorg.sh
+++ b/appvm-scripts/usrbin/qubes-run-xorg.sh
@@ -69,6 +69,7 @@ chmod 770 /var/run/xf86-qubes-socket
 
 export XDG_SEAT=seat0 XDG_VTNR=7 XDG_SESSION_CLASS=user
 
+<<<<<<< HEAD
 # Defaults value in case default-user value is not set
 if [ ! -z "$(qubesdb-read /default-user)" ];then
     DEFAULT_USER=$(qubesdb-read /default-user)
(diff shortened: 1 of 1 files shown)


yilmi:
Looks like debian buster build have been broken for some time now

appvm-scripts/usrbin/qubes-run-xorg.sh
```diff
@@ -69,4 +69,4 @@ chmod 770 /var/run/xf86-qubes-socket
 
 export XDG_SEAT=seat0 XDG_VTNR=7 XDG_SESSION_CLASS=user
 
-exec su -l user -c "/usr/bin/xinit $XSESSION -- $XORG :0 -nolisten tcp vt07 -wr -config xorg-qubes.conf > ~/.xsession-errors 2>&1"
+exec su -l $(qubesdb-read /default-user) -c "/usr/bin/xinit $XSESSION -- $XORG :0 -nolisten tcp vt07 -wr -config xorg-qubes.conf > ~/.xsession-errors 2>&1"
```
marmarek: This will break badly if user first update template and only then dom0. It should default to `user` if `/default-user` is missing.

appvm-scripts/usrbin/qubes-run-xorg.sh
```diff
@@ -69,4 +69,11 @@ chmod 770 /var/run/xf86-qubes-socket
 
 export XDG_SEAT=seat0 XDG_VTNR=7 XDG_SESSION_CLASS=user
 
-exec su -l user -c "/usr/bin/xinit $XSESSION -- $XORG :0 -nolisten tcp vt07 -wr -config xorg-qubes.conf > ~/.xsession-errors 2>&1"
+# Defaults value in case default-user value is not available
+if [ ! -z "$(qubesdb-read /default-user)" ];then
+    DEFAULT_USER=$(qubesdb-read /default-user)
```
marmarek: Negation of `-z` is `-n`.
Better save the value to some variable and use it instead of reading from qubesdb twice (what if only the second read fails?). You can even use `DEFAULT_USER` for that, and later set it to `user`, if its empty.

marmarek:
I'm going to merge this after final 4.0, so it will land as an update.

marmarek:
Can you squash those commits to cleanup git history?

PR #34, opened 2018-02-16
R4.1: Adapt to new gui protocol version using grant refs

HW42:


HW42:
Depends on https://github.com/QubesOS/qubes-gui-common/pull/2 and requires https://github.com/QubesOS/qubes-gui-daemon/pull/21 in dom0.

xf86-input-mfndev/configure.ac
```diff
@@ -63,13 +63,6 @@ AC_ARG_WITH([sdkdir], [],
     [sdkdir=`$PKG_CONFIG --variable=sdkdir xorg-server`])
 AC_SUBST([sdkdir])
 
-# Checks for libraries.
-AC_CHECK_LIB([u2mfn], [u2mfn_get_mfn_for_page])
-
-# Checks for header files.
-AC_HEADER_STDC
-AC_CHECK_HEADER([u2mfnlib.h])
-
```
marmarek: 🎉

gui-agent/vmside.c
```diff
@@ -260,30 +260,26 @@ void read_discarding(int fd, int size)
...
         exit(1);
     }
-    if (shmcmd.num_mfn == 0 || shmcmd.num_mfn > (unsigned)MAX_MFN_COUNT ||
-            shmcmd.width > MAX_WINDOW_WIDTH || shmcmd.height > MAX_WINDOW_HEIGHT) {
-        fprintf(stderr, "got num_mfn=0x%x for window 0x%x (%dx%d)\n",
-                shmcmd.num_mfn, (int) window, shmcmd.width, shmcmd.height);
-        read_discarding(g->xserver_fd,
-                shmcmd.num_mfn * sizeof(*mfnbuf));
+    if (wd_msg_len == 0) {
+        fprintf(stderr, "Failed to get window dump for window window 0x%lx\n",
```
cfcs: `window window`?

cfcs:
I think I recall @reynir also having problems with `gnttab_max_frames` (on the host side)? @reynir, do you remember the details?

@marmarek: would that mean it would be possible to run 8 windows at half size?

cfcs:
@marmarek that was the problem I was thinking of, thank you. :)

Re: Window size: 4 seems a bit limiting, at least I frequently have more than 4 full-screen windows open. If the default is to be conservative, any chance this could be made a configuration option?

marmarek:
> Re: Window size: 4 seems a bit limiting, at least I frequently have more than 4 full-screen windows open. If the default is to be conservative, any chance this could be made a configuration option?

Note that it's in practice about summary surface, expressed as number of
windows of maximum supported resolution, not just number of fullscreen
windows. It would be N (proposed 4) windows of _maximum supported
resolution_, which is currently 16384x6144. If you have lower
resolution, then you could have much more fullscreen (or else) windows.

If that limit will become an issue in the future, we could raise it
then, assuming more RAM will be a common thing at that time.

HW42:
Fixed the `domid != 0` case ;]

gui-domain-test

That's Xephyr (nested X server) running in a domain named `gui-domain` (with blue border from dom0). In `gui-domain` there's a `qubes-guid` running using the Xephyr server. This `qubes-guid` draws those green borders around the window from another domain. The terminal window itself doesn't get a border since there's no window manager running there which would need to draw the border.

cc @rootkovska

cfcs:
@marmarek Ahh, I thought it was of the current resolution. That constant definitely works for me. :)

xf86-qubes-common/Makefile
```diff
@@ -0,0 +1,10 @@
+XORG_SERVER_CFLAGS=$(shell pkg-config --cflags xorg-server)
+
+all: libxf86-qubes-common.so
+
+libxf86-qubes-common.so: xf86-qubes-common.c include/xf86-qubes-common.h
```
cfcs: shouldn't this have fstack-protector-all, ro-relocations etc?

gui-agent/vmside.c
```diff
@@ -258,32 +259,28 @@ void read_discarding(int fd, int size)
...
+                window);
         return;
     }
-    size = shmcmd.num_mfn * sizeof(*mfnbuf);
-    mfnbuf = alloca(size);
-    while (rcvd < size) {
-        ret =
-            read(g->xserver_fd, ((char *) mfnbuf) + rcvd,
-                    size - rcvd);
+    wd_msg_buf = alloca(wd_msg_len);
```
marmarek: Missing error handling (was already missing in the old version).

xf86-qubes-common/Makefile
marmarek: rpm already define them, it's just about using CFLAGS/LDFLAGS. I'll add it here too.

PR #41, opened 2018-05-26
Make it work with xorg-server 1.20 on Archlinux

kalkin:
Currently you can't upgrade Archlinux without skipping the xserver update. This PR fixes it. Seems to work for me.


marmarek:
Done in #42

PR #44, opened 2018-06-25
Add pulseaudio 12.0 header files

arjan-s:
This fixes the build at least on Archlinux, which is using 12.0 since a few days. Also, fix the install-time dependency on pulseaudio in the PKGBUILD.


na--:
Hey @arjan-s, thanks for this patch, just ran into this issue when building the new packages and was super happy that someone has already created a pull request!

I tested this and it mostly works - the package compiles and sound plays ok in the Archlinux VMs, but for some reason I can't attach the dom0 microphone. Attaching it doesn't outright fail, it seems to work - there are no error messages either through the GUI or via the dom0 CLI tools, and there's nothing in the VM or dom0 logs. But for some reason the microphone isn't actually usable from the Archlinux VM. The `Input Devices` tab in `pavucontrol` doesn't show anything:
pavucontrol-archlinux

While attaching the microphone to a fedora-28 VM works perfectly and it's usable there:
pavucontrol-fedora-28

So this seems to be something archlinux specific... Before the update, attaching the microphone to Archlinux VMs worked perfectly, but it's very possible that the issue isn't due to this pull request, I hadn't updated in a while... Do you also have the same problem?

btw here's how you can sign your commits so the status check doesn't fail:
- https://www.qubes-os.org/doc/code-signing/
- https://help.github.com/articles/signing-commits-using-gpg/

arjan-s:
Thank you for reporting this, I have the same issue. I have never used the microphone before, so that's why I didn't catch it. I'll try to find time to debug it and let you know. If you find the problem first, please let me know.

marmarek:
@arjan-s take a look at https://www.qubes-os.org/doc/code-signing/

PR #45, opened 2018-07-20
Add symlink from pulsecore-12.0 to pulsecore-12.2

virtualdxs:
Fixes issue where when building on fc26 for R4.0 would fail due to
pulsecore-12.2 headers being missing. ABI is identical between versions. 
Fixes QubesOS/qubes-issues#4109.


virtualdxs:
Damn it, I enabled signing on my build vm but not the vm I built this on. Let me fix that.

PR #48, opened 2018-08-20
archlinux: bump xorg-server version

ptitdoc:


virtualdxs:
Why do we require a specific version, so we have to bump it every time, as opposed to just requiring `xorg-server`?

marmarek:
The package require rebuild together with xorg-server (as it ships xorg modules). This way it will prevent xorg upgrade (which would break gui-agent) until gui-agent is rebuild with newer version.
Ideally it should be dynamically filled based on the xorg-server version used during build - we do that for Debian and Fedora. If you know Archlinux way for this, feel free to send PR.

PR #51, opened 2018-09-16
Fc29

fepitre:
QubesOS/qubes-issues#4223


debian/control
```diff
@@ -37,8 +37,8 @@ Depends:
     libxcomposite1,
     libxt6,
     libx11-6,
-    python,
-    python-xpyb,
+    python3,
+    python3-xpyb,
```
marmarek: This package doesn't exist. `xpyb` isn't python3 compatible, but its fork `xcffib` is. `python3-xcffib` is packaged [only for Debian sid (unstable)](https://packages.debian.org/search?keywords=python3-xcffib), but we can easily fix this.

fepitre: Should we package it along with the current one done for CentOS?

marmarek: Already done: https://github.com/QubesOS/qubes-python-xcffib/commit/e5d97334eb423773b7daff961c5fa8056e34537a

PR #60, opened 2018-12-30
Invoke the xsession scripts with bash so that .bash_profile etc are loaded

mig5:
This merge https://github.com/QubesOS/qubes-gui-agent-linux/commit/130865253fe03d37cdec17c8fabdae6bbcca1354 changed the way xinit is executed. It now invokes `/bin/sh`, which on Debian is 'dash' and not 'bash'.

The unforeseen effect is that config files like `.bash_profile` are not actually loaded when these X session startup scripts are run at boot.

As one example of this impact, `QUBES_GPG_AUTOACCEPT` user overrides in a GPG vault VM are not respected, because `.bash_profile` is ignored.

On Fedora it was probably unaffected since `/bin/sh` is still bash.

This commit forces `/bin/bash` which fixes QUBES_GPG_AUTOACCEPT (and potentially other things) for me.


mig5:
See comments in #4673

PR #66, opened 2019-06-10
add modargs for target domid and sink/source descr strings

xaki23:
this allows configuring audio vms (usb, bt) and connecting them as sinks/sources to dom0 pulseaudio.

`dom0$ pactl load-module module-vchan-sink domid=353 sink_desc=\'Qubes sink sys-bt\' source_desc=\'Qubes source sys-bt\'`
`sys-bt$ pacat-simple-vchan 0 dom0`

now you have a sink+source in dom0 mixer/pavucontrol, and dom0 in sys-bt mixer and can configure what goes where as needed.

building it is a mess at the moment since both parts are currently in very dom0/appvm specific qubes-gui packages.
only sane way i see is moving them into their own qubes-pulse component together and building it for both dom0 and distvms.

with that in place, adding a device class audio and a qvm-audio frontend to set up and tear down additional any-to-any audio connections is mostly trivial.

### DIFF
--- a/pulse/module-vchan-sink.c
+++ b/pulse/module-vchan-sink.c
@@ -85,16 +85,22 @@ PA_MODULE_DESCRIPTION("VCHAN sink/source");
 PA_MODULE_VERSION(PACKAGE_VERSION);
 PA_MODULE_LOAD_ONCE(false);
 PA_MODULE_USAGE("sink_name=<name for the sink> "
+		"sink_desc=<description string for the sink> "
 		"sink_properties=<properties for the sink> "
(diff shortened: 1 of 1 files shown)


marmarek:
> building it is a mess at the moment since both parts are currently in very dom0/appvm specific qubes-gui packages.

Actually, for R4.1 gui-daemon is built for VM too. The missing part is gui-agent for dom0, but it shouldn't be hard to add.

palensky:
I am on R4, and /usr/lib64/pulse-10.0/modules/module-vchan-sink.so is missing. 
How can i install it?

PR #67, opened 2019-06-17
Updated PKGBUILD to support xorg-server 1.20.5-1

Seres67:
Bump xorg-server version from 1.20.4 to 1.20.6 to support 1.20.5-1


marmarek:
I'll cherry-pick just the top commit to make git history cleaner. This change also looks ok to backport to R4.0, so I'll do that too.

PR #69, opened 2019-09-05
Add pulseaudio-12.99.2 symlinked headers

fepitre:


marmarek:
Backported to R4.0 already.

PR #73, opened 2019-09-24
Archlinux: Allow pulseaudio version between 13.0 and 14.0

neowutran:


neowutran:
( on a unrelated thing: I do not understand how the "code-signing" check is working. My GPG key is available on the key servers, ``gpg --recv-keys FB798BFB115895C9C95900649C2B29DD4A62ED27`` or ``gpg --recv-keys 0x0150AFFAFF83288F``, but the "code-signing" check end up with "Unable to verify (no valid key found)" )

marmarek:
>  I do not understand how the "code-signing" check is working.

Short answer: very unreliable, thanks to broken keyservers infrastructure.

marmarek:
In this case however, it looks like gpg version used by the bot doesn't support EDDSA signatures :/

marmarek:
A little update and it works now :)

neowutran:
Hehe, GG ! :)

PR #74, opened 2019-09-25
Archlinux: Allow any pulseaudio version between 13.0 and 14.0

neowutran:
for qubes-vm-pulseaudio


neowutran:
( same patch as yesterday, but for the branch release4.0 )

marmarek:
I've cherry-picked that patch to release4.0 already,

PR #77, opened 2019-11-14
qvm-features-request: add guivm-gui-agent

fepitre:


appvm-scripts/etc/qubes/post-install.d/20-qubes-guivm-gui-agent.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+if rpm -q xorg-x11-server-Xephyr > /dev/null; then
```
marmarek: Check binary existence, not the package. Otherwise it won't work on other distributions.

appvm-scripts/etc/qubes/post-install.d/20-qubes-guivm-gui-agent.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+if [ -e /usr/bin/Xephyr ]; then
+    qvm-features-request service supported-service.guivm-gui-agent=1
```
marmarek:
```suggestion
    qvm-features-request supported-service.guivm-gui-agent=1
```

PR #83, opened 2020-01-14
Allow GUI and AUDIO client to be ran before their corresonding servers

fepitre:
- gui-agent: put ExecPre in separate script
https://github.com/QubesOS/qubes-issues/issues/5662
https://github.com/QubesOS/qubes-issues/issues/833
https://github.com/QubesOS/qubes-issues/issues/1590


fepitre:
This will be improved by implementation of `qubesdb-read -w`.

PR #87, opened 2020-02-10
add ability to setup keymap with legacy /qubes-keyboard key

jevank:
As I understand there is new way to setup appVM keyboard with /keyboard-layout key and setxkbmap, but now it is broken in 4.1. Add addition code to fix it


jevank:
https://github.com/QubesOS/qubes-issues/issues/1396

jevank:
drop this temporary suggestion

PR #88, opened 2020-03-24
icon-sender: reconnect on failure

pwmarcz:
Currently, icon-sender breaks whenever the connection to icon-receiver
fails. This might happen if the user restarts the graphical session (for
example, logs out and logs in again).

Instead, whenever we're disconnected (qrexec-client-vm died), try to
reconnect. This is limited to one try to prevent thrashing.

The solution isn't perfect: after restarting the session, the existing
icons are all gone until you create a new window. This is because we
have no way of telling when icon-receiver is back until we try.

### DIFF
--- a/window-icon-updater/icon-sender
+++ b/window-icon-updater/icon-sender
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# -*- encoding: utf8 -*-
+# -*- encoding: utf-8 -*-
 #
 # The Qubes OS Project, http://www.qubes-os.org
 #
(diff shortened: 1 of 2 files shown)


marmarek:
Add some hold-off time (and maybe more retries?), otherwise logging of will most likely waste that one try for the time when user isn't logged in yet.

Alternatively, add waiting for user session in the receiver.

pwmarcz:
From what I see in manual testing, logging off and logging on again in dom0 is not noticed at all by icon-sender. I need a manual action (like creating a new window) to trigger a new call. So, no attempt is wasted, but the result looks a bit ugly. 

Ideally, I would have a "reconnect" procedure that waits until GUI is available, same as gui-agent does. But it's easy to do with vchan (just call `libvchan_*_init` and it will block until you're reconnected), and there is no equivalent mechanism with qrexec. Maybe there could be one?

pwmarcz:
See https://github.com/QubesOS/qubes-core-qrexec/pull/41, I want to do it in a different way using wait-for-sessino.

PR #91, opened 2020-03-28
vmside: update cursor position on LeaveNotify

pwmarcz:
When moving mouse through the window edge, the LeaveNotify event
contains cursor coordinates that are outside of window bounds.
Forward them to local window, otherwise the last event position
the window will receive will be from a MotionNotify still inside
the window.

This fixes an annoying glitch in Firefox: moving mouse cursor up
from a tab title (which touches the top edge of the window) makes
Firefox think we are still hovering over the tab, and displays a
tooltip, which stays there even after you switch to a different
program.

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -1551,6 +1551,10 @@ void handle_crossing(Ghandles * g, XID winid)
         XID window_under_pointer, root_returned;
         int root_x, root_y, win_x, win_y;
         unsigned int mask_return;
+
+        // move pointer out of window
+        feed_xdriver(g, 'M', attr.x + key.x, attr.y + key.y);
+
         ret =
             XQueryPointer(g->display, g->root_win, &root_returned,
                     &window_under_pointer, &root_x, &root_y,


pwmarcz:
Before:
Peek 2020-03-28 18-19

After:
Peek 2020-03-28 18-18

marmarek:
I can't reproduce the issue... Tried on fedora-30 and debian-10.
If you look at that function, there is a code to handle this case, by placing a tiny window just under the cursor.
The problem with just moving mouse away is that VM don't have full knowledge of all the windows (and also window decorations added by window manager in GUI VM/dom0). So, moving the mouse away may trigger EnterNotify in another window, while in fact real cursor is still over window decoration or a window of another VM. Take a look at https://github.com/QubesOS/qubes-issues/issues/368

Maybe there is an issue with that small window size/possition?

pwmarcz:
Okay, that is pretty weird: it breaks for me in Fedora 30 and 31, but not Debian 10.
Under both of these, the small window seems to be there, and judging by the cursor shape, it works:

image

(I added a 5x5 border, and had to use TightVNC instead of TigerVNC because the latter does not show mouse cursor)

The difference is Firefox version (Fedora has 74, Debian uses 68 ESR). I checked and Firefox 69 under Fedora also breaks. However, you're saying it works for you under Fedora 30, so this can't be it.

--

Also, when I run xev under Fedora in VM, I cannot see MotionNotify events. I can see them in Debian, and when running xev in dom0.

Any ideas?

marmarek:
I tested on Firefox 74. I also don't see MotionNotify on Fedora. I think GTK redirect those events to some other window (see `xwininfo -root -tree`), but haven't found exactly.

pwmarcz:
Okay, this is starting to look like a bug in Firefox: there are many of tooltip-related bugs on Mozilla's bug tracker, and I can make a variant of it happen also in dom0: if you switch to a different virtual desktop without moving mouse out of the window (using a keyboard shortcut), the tooltip remains on the screen. And this one also doesn't break in the Debian version. :)

Alright, I'm out of ideas. I tried some workarounds (like placing a stub window on new position, but it seems to be placed too late; or calling `XSync` after placing the stub window, but it breaks somehow).

Peek 2020-04-01 18-23
(Almost there but not quite - look at "Starred")

I think if there was a good way of ensuring the mouse is out of the window, it would be worth keeping, because (apart from fixing this) it makes the behaviour more like "normal" X. But for now I guess I'll keep just using this patch locally because Firefox is pretty unusable without it. Maybe I'll think of something later.

marmarek:
> (Almost there but not quite - look at "Starred")

What about it?

>  (like placing a stub window on new position, but it seems to be placed too late

Maybe it should be 3x3 (not 1x1) and placed so the cursor is in its cursor, not a corner? 

> if you switch to a different virtual desktop without moving mouse out of the window (using a keyboard shortcut), the tooltip remains on the screen.

I definitely see this issue from time to time, but cannot reliably reproduce it.

pwmarcz:
>>(Almost there but not quite - look at "Starred")
>What about it?

It changes color for a moment, which means that the cursor entered the second window for a moment, even though it was not supposed to (in Qubes' GUI it's still on the window title).

(Okay, I did not verify this situation with `xev` on the second window, maybe I should).

The code I was trying was someting like:

```diff
         if (window_under_pointer == winid) {
             XMoveResizeWindow(g->display, g->stub_win,
-                    root_x, root_y, 1, 1);
+                    attr.x + key.x, attr.y + key.y, 1, 1);
             XMapWindow(g->display, g->stub_win);
             XRaiseWindow(g->display, g->stub_win);
+            feed_xdriver(g, 'M', attr.x + key.x, attr.y + key.y);
         }
```

Looks like it should work, but apparently the window raises _after_ the mouse is moved, not before.

I tried to add `XSync()` just before `feed_xdriver` but somehow it did not work (the generated `LeaveNotify` events were wrong) - I don't know enough about X to understand what's happening here. Is `feed_xdriver` doing the same as `XWarpCursor` function, by the way?

>Maybe it should be 3x3 (not 1x1) and placed so the cursor is in its cursor, not a corner?

I can try that but based on the above I'm not sure how it would help. The position is right, it's a matter of order in which things happen.

PR #94, opened 2020-03-30
pulseaudio 13.99.1

pwmarcz:
Needs to be installed under LIBDIR/pulse-13.99.


Makefile
```diff
@@ -123,6 +124,8 @@ install-pulseaudio:
 		$(DESTDIR)/etc/pulse/qubes-default.pa
 	install -D pulse/module-vchan-sink.so \
 		$(DESTDIR)$(LIBDIR)/pulse-$(PA_VER)/modules/module-vchan-sink.so
+	install -D pulse/module-vchan-sink.so \
+		$(DESTDIR)$(LIBDIR)/pulse-$(PA_VER_SHORT)/modules/module-vchan-sink.so
```
marmarek: A symlink instead? Or make PA_VER always point to the right directory (check how pulseaudio calculate it)?

pwmarcz: Looking at pulseaudio source, the directory was always `pulse-{ver_major}.{ver_minor}` (I traced it as far back as 0.99.1). I updated the code to use that for directory.

PR #97, opened 2020-05-13
Do not run X server on sys-gui-gpu

cytadela8:
On sys-gui-gpu we use lightdm for running X server, but we still need
guivm-gui-agent service

Part of: QubesOS/qubes-issues#2618

### DIFF
--- a/appvm-scripts/qubes-gui-agent.service
+++ b/appvm-scripts/qubes-gui-agent.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Qubes GUI Agent
 After=qubes-meminfo-writer.service qubes-mount-dirs.service
+ConditionPathExists=!/var/run/qubes-service/lightdm
 
 [Service]
 StandardInput=tty


cytadela8:
Seems that the qubes-gui-agent service is started despite this condition

cytadela8:
Also reverting the other changes broke gui from vms

appvm-scripts/qubes-gui-agent.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Qubes GUI Agent
 After=qubes-meminfo-writer.service qubes-mount-dirs.service
+!ConditionPathExists=/var/run/qubes-service/lightdm
```
marmarek, line 4: Should be:
```suggestion
ConditionPathExists=!/var/run/qubes-service/lightdm
```

cytadela8: Thanks!!! I wouldn't have guested this is the problem.

cytadela8:
Tested.

PR #102, opened 2020-06-16
Debian guivm

fepitre:


appvm-scripts/usrbin/qubes-run-xephyr
```diff
@@ -2,4 +2,10 @@
 DISPLAY_XEPHYR="$1"
 OPTIONS_XEPHYR="-no-host-grab -fullscreen"
 
-LD_PRELOAD=/usr/lib64/qubes-gui-daemon/shmoverride.so /usr/bin/Xephyr $OPTIONS_XEPHYR "$DISPLAY_XEPHYR" > ~/.xephyr-errors 2>&1
+if [ -e /etc/debian_version ]; then
```
marmarek, line 5: Better check shmoverride.so existence directly.

fepitre: I adjusted with existence in both cases. In any other case, that will pass empty/env value.

marmarek:
When moving file to a different (sub) package you need to set correct dependencies, otherwise:
```
                        dpkg: error processing archive /tmp/apt-dpkg-install-qW6mV5/20-pulseaudio-qubes_4.1.8-1+deb10u1+devel1_amd64.deb (--unpack):
                         trying to overwrite '/usr/lib/tmpfiles.d/qubes-pulseaudio.conf', which is also in package qubes-gui-agent 4.1.8-1+deb10u1
```
See https://www.debian.org/doc/debian-policy/ch-relationships.html#overwriting-files-in-other-packages

PR #105, opened 2020-07-28
Extract Xephyr to a separate script

pwmarcz:
qubes-run-xorg will only start the "outer" Xorg instance. The
Xephyr session should be initiated using qrexec:

    qvm-run -p --no-gui --service $GUIVM qubes.GuiVMSession

The '-p' ensures that the above command will wait for the session
to finish.


pwmarcz:
WIP until I try it out with all the other parts.

PR #113, opened 2020-11-11
Start gui-agent only after systemd let in normal users

marmarek:
Use After=systemd-user-sessions.service
This way, we can be sure to start user applications (including those
from /etc/xdg/autostart) only when they are supposed to be started.
Other system services may order themselves before
systemd-user-session.service (or any of related .target unit).


marmarek:
PipelineRetry

PR #117, opened 2020-11-26
Fix Unicode handling in Qubes clipboard

DemiMarie:
libX11 claims to have Xutf8TextListToTextProperty for handling UTF-8
text, but it is just broken and winds up mangling the text instead.
Instead, use XStringListToTextProperty and validate UTF-8 validity
ourselves.

Fixes https://github.com/QubesOS/qubes-issues/issues/2845

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -49,6 +49,7 @@
 #include "txrx.h"
 #include "list.h"
 #include "error.h"
+#include "encoding.h"
 #include <libvchan.h>
 
 #define SOCKET_ADDRESS  "/var/run/xf86-qubes-socket"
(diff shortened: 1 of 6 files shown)


marmarek (changes requested):
Please split adding 'static' to a separate commit.

gui-agent/vmside.c
```diff
@@ -1041,21 +1044,31 @@ void process_xevent_selection_req(Ghandles * g,
...
+    else if (req->target == Compound_text)
         convert_style = XCompoundTextStyle;
+    else if (req->target == Utf8_string_atom)
+        convert_style = XUTF8StringStyle;
     if (convert_style != XConverterNotFound) {
         XTextProperty ct;
-        Xutf8TextListToTextProperty(g->display,
-                (char **) &g->clipboard_data,
-                1, convert_style, &ct);
+        char **ptr = { NULL };
```
marmarek, line 1055: This should be `char *ptr[0]`, otherwise it points at an initializer in rodata section.

And indeed, when pasting characters from the very issue title it tries to fix, I got SEGV.

We do have some tests for inter-VM clipboard (https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/tests/integ/basic.py#L468), maybe we need one for weird unicode characters too? (but that isn't blocking for this PR)

marmarek (changes requested):

marmarek (approved):

PR #121, opened 2020-12-09
Implement WM_STATE in X11 GUI agent

bemoody:
These changes implement the WM_STATE property on the X11 AppVM side - allowing applications to identify the toplevel windows of other applications that are running in the same VM.

This is necessary for "screen sharing" of individual windows in Firefox and possibly other, similar programs: https://github.com/QubesOS/qubes-issues/issues/5789

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -89,7 +89,7 @@ struct _global_handles {
     Atom tray_opcode;      /* Atom: _NET_SYSTEM_TRAY_MESSAGE_OPCODE */
     Atom xembed_info;      /* Atom: _XEMBED_INFO */
     Atom utf8_string_atom; /* Atom: UTF8_STRING */
-    Atom wm_state;         /* Atom: _NET_WM_STATE */
+    Atom net_wm_state;     /* Atom: _NET_WM_STATE */
     Atom wm_state_fullscreen; /* Atom: _NET_WM_STATE_FULLSCREEN */
     Atom wm_state_demands_attention; /* Atom: _NET_WM_STATE_DEMANDS_ATTENTION */
     Atom wm_take_focus;    /* Atom: WM_TAKE_FOCUS */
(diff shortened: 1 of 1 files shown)


bemoody:
V2: use the correct variable name (winid -> window), and added gpg signatures.  Sorry I accidentally pushed a broken version the first time.

On a side note, I found that trying to build this using `dpkg-buildpackage -b` fails, because it expects the variable BACKEND_VMM to be defined.  That variable presumably ought to be defined either in the project makefile or the debian/rules file.

marmarek (changes requested):
Please rebase it (and make the PR) on master branch. When merged, I'll cherry pick it to release4.0.
Otherwise looks good.

bemoody:
Now rebased and I think I set the target branch correctly.

I haven't *tested* this on 4.1, but it ought to work. :)

DemiMarie:
@bemoody have you uploaded your key to a keyserver?  If not, the pipeline probably just needs to be retried.

marmarek:
PipelineRetry

marmarek:
(the above is about gitlab, not signature check - the key is still missing)

bemoody:
My public key ID is FBE2DC9A105D63ABCC5E679D323A63063AC869DF and it
should be on the keyservers.  Let me know if I need to upload it
anywhere in particular, or anything else I can do to verify myself. ;)

bemoody:
Oh, I used a signing subkey, so perhaps that's confusing the
verification system?  My subkey ID is
F0D84E04D8D1935EE37A14E9905D8AB782223619.

DemiMarie:
Using a subkey is considered best practice, and it should not confuse the verification scripts.  If it does, that’s a bug in the scripts.  @marmarek have the scripts been tested with a subkey?

marmarek:
> My public key ID is FBE2DC9A105D63ABCC5E679D323A63063AC869DF and it should be on the keyservers

```
$ gpg --keyserver pool.sks-keyservers.net  --recv-key FBE2DC9A105D63ABCC5E679D323A63063AC869DF
gpg: keyserver receive failed: No data
```
Which one?

marmarek:
PipelineRetry

DemiMarie:
> > My public key ID is FBE2DC9A105D63ABCC5E679D323A63063AC869DF and it should be on the keyservers
> 
> ```
> $ gpg --keyserver pool.sks-keyservers.net  --recv-key FBE2DC9A105D63ABCC5E679D323A63063AC869DF
> gpg: keyserver receive failed: No data
> ```
> 
> Which one?

```
gpg --recv-keys --keyserver=keys.openpgp.org -- FBE2DC9A105D63ABCC5E679D323A63063AC869DF
```

works.

marmarek (approved):
Thanks! It works!

PR #122, opened 2020-12-11
Ensure we call XSetInputFocus NAND send WM_TAKE_FOCUS

DemiMarie:
Based on i3 commit 21a2971b2442ab0881cf79553cc6b65bbb44afa7

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -1862,14 +1862,14 @@ static void handle_focus(Ghandles * g, XID winid)
             use_take_focus = False;
         }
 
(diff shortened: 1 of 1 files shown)


marmarek:
Hmm, the change you just pushed is restoring a possibility to use both focus switch methods, just in a reversed order. Is that intentional? I though this caused the issue, no?

DemiMarie:
@marmarek I honestly have no idea right now.  I will need to test it.

DemiMarie:
Closing in favor of #124, which actually fixes the problem.

PR #124, opened 2020-12-17
Fix globally active clients

DemiMarie:
Globally active clients will unfocus themselves.  Don’t tell them their window is no longer focused.

Fixes https://github.com/QubesOS/qubes-issues/issues/3982.


marmarek:
Re version change - you may want to set `INCREMENT_DEVEL_VERSIONS=1` in your builder.conf to get version (release number) incremented each build - makes it easier to deploy on the test system.

gui-agent/vmside.c
```diff
@@ -964,53 +951,50 @@ static void handle_targets_list(Ghandles * g, Atom Qprop, unsigned char *data,
...
     }
-    XConvertSelection(g->display, Clp,
-            have_utf8 ? g->utf8_string_atom : XA_STRING, Qprop,
-            g->stub_win, CurrentTime);
+    XConvertSelection(g->display, g->clipboard,
+            have_utf8 ? g->utf8_string_atom : XA_STRING, g->qprop,
+            g->stub_win, time);
 }
 
+_Thread_local Time the_time;
```
marmarek, line 959: gui-agent is a single-thread application, can it be in `Ghandles` instead? In any case, declare global stuff at the beginning of the file.

gui-agent/vmside.c
```diff
@@ -876,7 +865,6 @@ static void process_xevent_configure(Ghandles * g, XID window,
     struct msg_hdr hdr;
     struct msg_configure conf;
     struct genlist *l;
-    /* SKIP_NONMANAGED_WINDOW; */
```
marmarek, line 879: This one was kind of intentional, as a hint the below fragment is open-coded version of this macro. Maybe too subtle one...

marmarek (changes requested):

gui-agent/vmside.c
marmarek: With this change, you don't need extra parameter anymore (functions do receive `g` reference already).

marmarek:
PipelineRetry

marmarek (approved):

PR #125, opened 2020-12-17
Avoid passing NULL as an X11 string

DemiMarie:
Xlib helpfully doesn’t dereference the NULL pointer, but this isn’t
documented.  Also add a comment that explains why we don’t immediately
return if the UTF-8 validity check fails.


marmarek:
PipelineRetry

marmarek (approved):

PR #139, opened 2021-07-06
Always keep latin keyboard layout

jevank:
Setting only non-latin layout breaks hotkeys in various applications, so setup secondary 'us' to propagated from dom0.

Fixes https://github.com/QubesOS/qubes-issues/issues/5431
Fixes https://github.com/QubesOS/qubes-issues/issues/6690

### DIFF
--- a/appvm-scripts/usr/lib/qubes/qubes-keymap.sh
+++ b/appvm-scripts/usr/lib/qubes/qubes-keymap.sh
@@ -11,6 +11,10 @@ set_keyboard_layout() {
     KEYMAP_LAYOUT="$(echo "$KEYMAP"+ | cut -f 1 -d +)"
     KEYMAP_VARIANT="$(echo "$KEYMAP"+ | cut -f 2 -d +)"
     KEYMAP_OPTIONS="$(echo "$KEYMAP"+ | cut -f 3 -d +)"
+    if [ "$KEYMAP_LAYOUT" != "us" ]; then
+        KEYMAP_LAYOUT="$KEYMAP_LAYOUT,us"
+    fi
+
     if [ -n "$KEYMAP_VARIANT" ]; then
         KEYMAP_VARIANT="-variant $KEYMAP_VARIANT"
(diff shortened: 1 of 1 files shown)


marmarek:
PipelineRefresh

marmarek (approved):

solardiz:
This addition of `,us` layout breaks the `si` layout for me. This appears to help, and doesn't appear to break the shortcut keys:

```shell
+++ /usr/lib/qubes/qubes-keymap.sh	2023-07-09 15:00:38.085000000 +0200
@@ -11,7 +11,7 @@
     KEYMAP_LAYOUT="$(echo "$KEYMAP"+ | cut -f 1 -d +)"
     KEYMAP_VARIANT="$(echo "$KEYMAP"+ | cut -f 2 -d +)"
     KEYMAP_OPTIONS="$(echo "$KEYMAP"+ | cut -f 3 -d +)"
-    if [ "$KEYMAP_LAYOUT" != "us" ]; then
+    if [ "$KEYMAP_LAYOUT" != "us" -a "$KEYMAP_LAYOUT" != "si" ]; then
         KEYMAP_LAYOUT="$KEYMAP_LAYOUT,us"
         KEYMAP_VARIANT="$KEYMAP_VARIANT,"
     fi
```

I guess the problem isn't limited to just this one layout, so ideally we'd figure it out and fix it in a general manner, but meanwhile maybe we should get the above hack in?

marmarek:
> This addition of `,us` layout breaks the `si` layout for me

What exactly is broken then?

solardiz:
> > This addition of `,us` layout breaks the `si` layout for me
> 
> What exactly is broken then?

WIth `si,us`, I am getting behavior same or similar to that for `us` alone - I cannot type characters specific to the `si` layout. With `si` alone, I can.

Somehow the same problem does not occur for `ru,us`, which appears to work just like `ru` alone.

I tested all of these with a Debian 10 VM.

PR #151, opened 2022-02-23
Make qubes-session inherit the systemd user session variables.

Rudd-O:
The standard environment.d facility is currently not respected in Qubes OS 4.1.  This leads to inability to define environment variables when starting apps via qrexec¹.

This should be fixed.  Here is the fix — it is a fix that will not have any substantial performance impact (it runs *once*), and uses documented APIs to execute on the fix.  Furthermore, the fix enables packaged software to ship environment variables in `/usr/lib/environment.d` as per the standard, without which presumable at least some software will simply be broken.

¹ Technically there's `/etc/profile.d` but that's meant to be used for **shells**.  There is no guarantee that apps started via qrexec will inherit these variables, and that doesn't cover software packaging factory environment variable customization in `/usr/lib/environment.d`.

### DIFF
--- a/appvm-scripts/usrbin/qubes-session
+++ b/appvm-scripts/usrbin/qubes-session
@@ -27,6 +27,14 @@
 
 loginctl activate "$XDG_SESSION_ID"
 
+# Now import the environment from the systemd user session.
+# This is necessary to enable users to configure their
+# Qubes environment using the standard environment.d
+# facility.  Documentation for the facility is at:
+# https://www.freedesktop.org/software/systemd/man/environment.d.html
+eval `systemctl --user show-environment | sed 's/^/export /'`
+
+
 if qsvc guivm-gui-agent; then
     if [ -e "$HOME/.xinitrc" ]; then
         . "$HOME/.xinitrc"
--- a/appvm-scripts/usrbin/qubes-session
+++ b/appvm-scripts/usrbin/qubes-session
@@ -32,7 +32,7 @@ loginctl activate "$XDG_SESSION_ID"
 # Qubes environment using the standard environment.d
 # facility.  Documentation for the facility is at:
 # https://www.freedesktop.org/software/systemd/man/environment.d.html
-eval `systemctl --user show-environment | sed 's/^/export /'`
+eval $(systemctl --user show-environment | sed 's/^/export /')
 
 
 if qsvc guivm-gui-agent; then


jevank:
I think better way is to enable pam_env.so 
/etc/pam.d/qubes-gui-agent:
```
...
session         required        pam_env.so
```

Rudd-O:
pam_env does not implement environment.d

https://man7.org/linux/man-pages/man8/pam_env.8.html

We may want to add that **too**, but that is a different enhancement request.

jevank:
Ouch, this is systemd routine which include /etc/environment.

We use pam_env.so to setup env per qube via /home/user/.environment and global via /etc/environment.

Rudd-O:
> We use pam_env.so to setup env per qube via /home/user/.environment and global via /etc/environment.

Not even sure that actually works, but I didn't try it.  All I know is that the systemd user session implements `environment.d`, but it's 100% not clear to me how *actual desktop sessions* actually are meant to incorporate those variables in their environment.  Given the absence of the environment variables set up via systemd user session, in the context of qubes-session, it's clear that qubes-session itself is not running within the systemd session.  So I am using this eval trick to work around that.

If there is a technically more sound way of implementing it, then we should use that.

Rudd-O:
```
[user@projects ~]$ loginctl show-session c2
Id=c2
User=1000
Name=user
Timestamp=Wed 2022-02-23 12:07:59 UTC
TimestampMonotonic=4458231
VTNr=7
Seat=seat0
TTY=tty7
Display=:0
Remote=no
Service=qubes-gui-agent
Scope=session-c2.scope
Leader=542
Audit=4294967295
Type=x11
Class=user
Active=yes
State=active
IdleHint=no
IdleSinceHint=0
IdleSinceHintMonotonic=0
LockedHint=no
```

Actually...

```
root         528  0.0  0.0   5008  3704 tty7     Ss+  12:07   0:05 /usr/bin/qubes-gui -d 0
root         542  0.0  0.1  13200  6908 tty7     S+   12:07   0:00  \_ /usr/bin/qubes-gui-runuser user /bin/sh -l -c exec /usr/bin/xinit /etc/X11/xinit/xinitrc -- /usr/libexec/Xorg :0 -nolisten tcp vt07 -wr -config xorg-qubes.conf > ~/.xsession-errors 2>&1
user         564  0.0  0.0   3924  2636 ?        Ss   12:07   0:00      \_ /usr/bin/xinit /etc/X11/xinit/xinitrc -- /usr/libexec/Xorg :0 -nolisten tcp vt07 -wr -config xorg-qubes.conf
user         667  0.1  3.5 969792 207508 ?       Sl   12:07   0:43          \_ /usr/libexec/Xorg :0 -nolisten tcp vt07 -wr -config xorg-qubes.conf
user         683  0.0  0.0  11464  3532 ?        Ss   12:07   0:00          \_ /usr/bin/bash /usr/bin/qubes-session
user         735  0.0  0.0   7200   524 ?        Ss   12:07   0:00              \_ /usr/bin/ssh-agent /etc/X11/xinit/Xclients
user         814  0.0  0.0   9632   808 ?        S    12:07   0:00              \_ sleep inf
```

The programs started by `qrexec-fork-server` aren't under the influence of the `qubes-session` program.  That means this fix is necessary, but incomplete.

Rudd-O:
Yeah qrexec-fork-server is going to need the same treatment.  It's not clear to me what even starts it.

marmarek:
> It's not clear to me what even starts it.

It is started from qubes-session, you don't see it on the process tree because its parent exits in the meantime.

Rudd-O:
AH!  Then this fix is complete.  What needs to be done before merging it?

appvm-scripts/usrbin/qubes-session
```diff
@@ -27,6 +27,14 @@
 
 loginctl activate "$XDG_SESSION_ID"
 
+# Now import the environment from the systemd user session.
+# This is necessary to enable users to configure their
+# Qubes environment using the standard environment.d
+# facility.  Documentation for the facility is at:
+# https://www.freedesktop.org/software/systemd/man/environment.d.html
+eval `systemctl --user show-environment | sed 's/^/export /'`
```
fepitre, line 35: Please use `$(...)` instead of the deprecated syntax.

Rudd-O: Done.

marmarek (approved):

Rudd-O:
https://rudd-o.com/linux-and-free-software/setting-up-environments-variables-on-specific-qubes-os-qubes

fepitre:
@Rudd-O thank you for the feedback. @andrewdavidwong we may add this tip somewhere in the doc?

Rudd-O:
I'm thinking of a `qvm-envvar` command like `qvm-prefs`.  Probably overkill, ignore me.

andrewdavidwong:
> @Rudd-O thank you for the feedback. @andrewdavidwong we may add this tip somewhere in the doc?

We welcome the addition of good external documentation links to the [external documentation section](https://www.qubes-os.org/doc/#external-documentation) of the table of contents. To add or update a link, please feel free to [create a pull request against the documentation index](https://www.qubes-os.org/doc/how-to-edit-the-documentation/#how-to-edit-the-documentation-index).

PR #153, opened 2022-03-14
WIP: Make window focus work well enough

locriacyber:
This is a hack, because X is poorly designed to be able to mirror focus from another X server. Close https://github.com/QubesOS/qubes-issues/issues/7345.
This is a fork from my `mod-keys` branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,14 @@ tags
 deb/
 pkgs
 debian/changelog.*
+/PKGBUILD*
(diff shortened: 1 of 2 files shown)


locriacyber:
This is a failure. The original branch (`master`) sometimes has more focus than perfect, while this branch sometimes has less focus than perfect. The original implementation is buggy, but better. More focus ≠ more key events. Less focus = dropped key events.

PR #154, opened 2022-04-01
Fix focus handling (attempt 2)

locriacyber:
This resolves most focus-related issues.

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -1671,6 +1671,8 @@ static void handle_keypress(Ghandles * g, XID UNUSED(winid))
     if (!is_mod_key) feed_xdriver(g, 'K', key.keycode, is_press);
 }
 
+static void handle_focus_helper(Ghandles * g, XID winid, struct msg_focus msg);
+
 static void handle_button(Ghandles * g, XID winid)
 {
     struct msg_button msg;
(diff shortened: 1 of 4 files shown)


locriacyber:
The client side is working. The server side still need some work.

locriacyber:
I hope this is the last fix about window focus before wayland.

locriacyber:
Before [1c5b880](https://github.com/QubesOS/qubes-gui-agent-linux/pull/154/commits/1c5b880ce7158517a0337b777231c0987d5f6db8): consistently buggy (when VM intra-transfer focus)

After: sometimes window will lose keyboard focus (mouse works fine). Minimize->maximize won't fix the problem. Have to click desktop when window is showing to fix.

locriacyber:
Remaining problems:

- Release all keys pressed after focus out of VM, like, 1 second after. By the time switched back, the VM should think all keys are released. Example: when VM think Ctrl + D is stuck, it close terminal immediately after you open one.
- Decide which window has decoration, that we should give focus implicitly.
Focusing every X "window" on click is working... mostly, except for VS Code's context menu, or Firefox's context menu. Leafpad's context menu works fine though. You cannot click on context menu with that attempt, so I didn't add this behaviour.

gui-agent/vmside.c
```diff
@@ -1679,19 +1692,43 @@ static void handle_button(Ghandles * g, XID winid)
...
+    // Fake a "focus in" when mouse down on unfocused window.
+    if (is_button_press) {
+        // int _return_to;
+        // XID focused_winid;
+        // XGetInputFocus(g->display, &focused_winid, &_return_to);
+        // bool a = focused_winid != winid
+        // TODO: check if window is "top-level" (has border), and only focus if it or its child was not focused
+        // some applications get confused if you give focus to its context menu
+        // I'm not sure how this works yet
+        bool need_focus = false;
```
locriacyber, line 1712: If anyone know how to do this properly, please send pull request to my branch (`fix-focus-2`)

DemiMarie: Windows have decorations if and only if they are not override-redirect.  Context menus can just be not given focus.

DemiMarie:
@locriacyber Would you mind splitting the PKGBUILD changes into a separate PR?

> Remaining problems:
> 
> 1. Release all keys pressed after none of a VM's window is focused, like, 1 second after. (what keys are pressed could be tracked by dom0) By the time switched back, the VM should think all keys are released. Example: when VM think Ctrl + D is stuck, it close terminal immediately after you open one.
> 2. Decide which window has decoration, that we should give focus implicitly.
>    Focusing every X "window" on click is working... mostly, except for VS Code's context menu, or Firefox's context menu. Leafpad's context menu works fine though. You cannot click on context menu with that attempt, so I didn't add this behaviour.

Are these regressions, or are they problems with the current agent too?

locriacyber:
They are new problems (recalling from memory, since I haven't use old agent for awhile).

Problem 1: Before this patch, key repeat was disabled in AppVM. Even if a key was stuck in AppVM, it won't generate stuff likeeeeeeeeeeeeeeeeee <- this.
Problem 2: I remember the old xside + vmside code send focus event to window on every click. I browsed the git history, but couldn't find a trace of the behavior "focus window on click".

locriacyber:
Superseded by https://github.com/QubesOS/qubes-gui-agent-linux/pull/169

PR #157, opened 2022-04-29
Pipewire agent

DemiMarie:
This now actually works.

Fixes https://github.com/QubesOS/qubes-issues/issues/6358

Edit: This builds ordinary Fedora/Debian/Arch packages.  To use them, install them in your VM of choice, then enable PipeWire with `qvm-service somevm pipewire on`.

Edit 2: I have fixed all the bugs that I am aware of and able to find.  ~~There is still one remaining bug: if I am recording an audio stream, and start playback _while the recording is in progress_, audio playback may stop working.  However, I believe this is _not_ an agent bug, but rather a bug in pacat-simple-vchan, as I am also able to get pacat-simple-vchan to behave strangely in other ways (such as not responding to control requests while recording silence).~~ Now fixed I think.

Edit 3: All remaining bugs appear to be fixed, at least if `pacat-simple-vchan` has QubesOS/qubes-gui-daemon#109.

Edit 4:  Remaining to-do items (should not block merge):

- [ ] Allow some properties (such as buffer sizes) properties to be configured via QubesDB as well as the configuration file.
- [ ] Watch `/qubes-audio-domain-xid` for changes.  If it does change, disconnect from the old audio VM and connect to the new one.
- [ ] Provide an interface that an audio VM can use to differentiate between a microphone being muted and disconnected.  Useful for workloads that care about this distinction, if any.
- [ ] Allow the node name to be configured via the configuration file.
- [ ] Integration with qubes-video-companion for video support.
- [ ] Implement buffer negotiation as suggested in the header file.


marmarek:
PipelineRefresh

DemiMarie:
PipelineRetry

DemiMarie:
PipelineRefresh

DemiMarie:
PipelineRetry

pipewire/GNUmakefile
```diff
@@ -0,0 +1,17 @@
+MAKEFLAGS ::= -rR
+override CC ::= $(or $(CC),gcc)
+override CFLAGS ::= $(CFLAGS) $(shell pkg-config --cflags libpipewire-0.3 vchan-xen) -fno-strict-aliasing -fno-strict-overflow -fno-delete-null-pointer-checks -fPIC -Wall -Wextra -Werror -Wno-missing-field-initializers -Wno-unused-parameter -ggdb -O2 -D_FORTIFY_SOURCE=2
+override LDLIBS ::= $(LDFLAGS) $(shell pkg-config --libs libpipewire-0.3 vchan-xen) -lqubesdb -ggdb
```
marmarek, line 4: Please, don't use `override` (or other ways to override explicitly requested build options). This should be `CC ?= gcc`, and `..FLAGS += ...`.

DemiMarie: I dropped the use of `override` for `CFLAGS` and `LDLIBS`, but doing so for `CC` ran into a GNU make bug (`CC` got set to `cc`, not `gcc`).  Therefore, `override` is still used for `CC`, but if `CC` is set in any way other than by a GNU make default, a conditional will choose the existing value.

marmarek: It isn't GNU make bug, it's a bug in your makefile (or your assumptions). `cc` is a perfectly valid value for `CC` variable.

DemiMarie: `cc` _is_ a valid value for `CC`.  The bug is that I used `MAKEFLAGS ::= -rR` to tell GNU make _not_ to set any implicit variables, but it sets this variable anyway.  That said, `cc` is probably a better default than `gcc`, as there are platforms where `gcc` is _not_ the system C compiler.

marmarek: > I used `MAKEFLAGS ::= -rR`

Okay, that's another thing that shouldn't be there. I'm not sure why you hate those defaults so much (and I do _not_ want to know), but please stop.

marmarek: Oh, and also: please use _standard_ `Makefile` name, not `GNUmakefile`.

DemiMarie:
At this point, the only code that I _know_ to be missing is the Debian and Arch maintainer scripts to manage PipeWire-related systemd units.

marmarek (changes requested):

pipewire/qubes-pw-module.c
```diff
@@ -0,0 +1,1148 @@
...
+        if (playback_size &&
+            (res = parse_number(playback_size, INT32_MAX / 2, &write_min, "playback buffer size")))
+            goto error;
+
+        impl->stream[PW_DIRECTION_OUTPUT].buffer_size = read_min;
+        impl->stream[PW_DIRECTION_INPUT].buffer_size = write_min;
+    }
+
+    pw_log_error("module %p: new (%s), peer id is %d", impl, args, (int)impl->domid);
+    pw_log_error("module %p: record buffer size %zu, playback buffer size %zu",
```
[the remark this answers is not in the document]
DemiMarie: I chose to log it as “info” because someone might very well want to know what the buffer size they had was, and debug logging is off by default.

DemiMarie (commented):
Comments (hopefully) addressed.  I will not be able to re-test until tomorrow.

debian/qubes-vm-pipewire.postinst
```diff
@@ -0,0 +1 @@
+systemctl --global --no-reload preset wireplumber.service pipewire.service pipewire.socket || :
```
marmarek, line 1: For any custom maintainer script, please include `#DEBHELPER#` placeholder auto automatically generated parts (see other packages). In fact, you may be able to automatically generate this one too, using [dh_installsystemduser](https://www.man7.org/linux/man-pages/man1/dh_installsystemduser.1.html) - it's enabled by default, but you may need to tell it which units to handle, as you install just drop-ins here, not the actual units.

marmarek (changes requested):
Some packaging comments

pipewire/qubes-pw-module.c
```diff
@@ -1061,23 +1074,27 @@ int pipewire__module_init(struct pw_impl_module *module, const char *args)
...
 
-        size_t read_min = 1 << 15, write_min = 1 << 15;
+        size_t read_min = SIZE_MAX, write_min = SIZE_MAX;
         const char *record_size = pw_properties_get(props, QUBES_PW_KEY_RECORD_BUFFER_SPACE);
         const char *playback_size = pw_properties_get(props, QUBES_PW_KEY_PLAYBACK_BUFFER_SPACE);
         if (!record_size || !playback_size) {
             const char *buffer_size = pw_properties_get(props, QUBES_PW_KEY_BUFFER_SPACE);
-            if (buffer_size && !record_size)
+            if (!buffer_size) { // should always be set
+                pw_log_error("Packaging bug: %s not specified in configuration file",
```
marmarek, line 1089: I don't think that's a good idea, better to always have some built-in default.

DemiMarie: Will do, especially since PipeWire makes it difficult for a configuration file’s default to be overridden by the user, yet still get picked up by this module.

marmarek:
Some quick test results (this PR + stock pacat-simple-vchan):
 - playback seems to work, both via pipewire natively and via pipewire-pulseaudio
 - recording seems to work with pipewire natively, but is very choppy via pipewire-pulseaudio
 - the module seems to properly sent record start/stop commands
 - the module seems to _not_ send playback start/stop commands, instead it sends silence (zeroes) when nothing is playing -> this prevents the actual device to suspend when unused, and results in constant CPU usage on the pacat-simple-vchan side

marmarek:
- [ ] when pipewire is started first, then some client (trying to) play something and only then `pacat-simple-vchan`, then log has endless stream of `Overrun: asled tp wrote 2048 bytes, but can only write 0`
- [ ] when recording is started in VM first, but recording is allowed later, there is a stream of `Underrun: asked to read 2048 bytes, but only xx available` (where `xx` is mostly non-zero, but sometimes 0). At ~one case~ two cases, when I stopped recording in the VM (closed the client), then pipewire crashed with SIGABRT and `corrupted size vs prev_size` message (I have a backtrace and coredump, but I'm not sure how useful they will be in practice)

In both of the above, the client is really connected via pipewire-pulseaudio (pavucontrol and/or firefox playing something)

DemiMarie:
> Some quick test results (this PR + stock pacat-simple-vchan):

> * [ ]  recording seems to work with pipewire natively, but is very choppy via pipewire-pulseaudio

Will need to debug this.  Does changing the max quantum in `30_qubes.conf` help?

> * [ ]  the module seems to _not_ send playback start/stop commands, instead it sends silence (zeroes) when nothing is playing -> this prevents the actual device to suspend when unused, and results in constant CPU usage on the pacat-simple-vchan side

I am pretty sure that these _do_ get sent.  What does _not_ get sent is an initial cork command at startup.  I can add that easily, but, pavucontrol seems to keep an audio stream open all the time.  `libpipewire-module-qubes.so` is just relaying the events PipeWire provides it, so I think this part is a bug in either PipeWire or pavucontrol.

> * [ ]  when pipewire is started first, then some client (trying to) play something and only then `pacat-simple-vchan`, then log has endless stream of `Overrun: asled tp wrote 2048 bytes, but can only write 0`

I have not managed to reproduce this yet.

> * [ ]  when recording is started in VM first, but recording is allowed later, there is a stream of `Underrun: asked to read 2048 bytes, but only xx available` (where `xx` is mostly non-zero, but sometimes 0). At ~one case~ two cases, when I stopped recording in the VM (closed the client), then pipewire crashed with SIGABRT and `corrupted size vs prev_size` message (I have a backtrace and coredump, but I'm not sure how useful they will be in practice)

Please send both to me privately.

DemiMarie:
> * [ ]  the module seems to _not_ send playback start/stop commands, instead it sends silence (zeroes) when nothing is playing -> this prevents the actual device to suspend when unused, and results in constant CPU usage on the pacat-simple-vchan side

Fixed by reporting the current stream state on all vchan connections.

> * [ ] when pipewire is started first, then some client (trying to) play something and only then `pacat-simple-vchan`, then log has endless stream of `Overrun: asled tp wrote 2048 bytes, but can only write 0`

Hopefully this is fixed as well.

DemiMarie:
> * [ ]  when recording is started in VM first, but recording is allowed later, there is a stream of `Underrun: asked to read 2048 bytes, but only xx available` (where `xx` is mostly non-zero, but sometimes 0). At ~one case~ two cases, when I stopped recording in the VM (closed the client), then pipewire crashed with SIGABRT and `corrupted size vs prev_size` message (I have a backtrace and coredump, but I'm not sure how useful they will be in practice)

I get occasional underruns, but they stop after a few seconds.  No crashes so far.

marmarek:
CI:
```
qubes-pw-module.c: In function 'playback_stream_process':
qubes-pw-module.c:643:22: error: unused variable 'control_vchan' [-Werror=unused-variable]
  643 |     struct libvchan *control_vchan = capture_stream->vchan;
      |                      ^~~~~~~~~~~~~
```

DemiMarie:
PipelineRetryFailed

marmarek (changes requested):

marmarek:
Normally I'd ask to squash those commits into a smaller set, but in this particular case, commit messages are elaborate enough to worth keeping them even if some commits fix things that were buggy within this very PR.
But I do have one request here: please eliminate the one pair of commit+revert, and squash the one `fixup!` into commit pointed by it.

DemiMarie:
PipelineRetryFailed

DemiMarie:
PipelineRetryFailed

archlinux/PKGBUILD-pipewire-qubes.install
```diff
@@ -0,0 +1,15 @@
+## arg 1:  the new package version
+post_install() {
+  systemctl --global --no-reload preset pipewire.service pipewire.socket wireplumber.service || :
+  qvm-features-request supported-service.pipewire=1
```
marmarek, line 4: This is not the way to do it (nor in other package types). Create a file in `/etc/qubes/post-install.d` for it. See https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-features.html#services or existing scripts there for details.

marmarek (changes requested):

archlinux/PKGBUILD-pipewire-qubes.install
marmarek: In this repo, there is already `appvm-scripts/etc/qubes/post-install.d` for it.

marmarek:
https://openqa.qubes-os.org/tests/69017

I'm not sure if the issue is with tests, or the agent, but looks more likely to be an agent issue, so commenting here:
1. Pipewire doesn't start on Debian:
```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 149, in common_audio_playback
    self.loop.run_until_complete(self.testvm1.run_for_stdio(cmd))
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qubes/vm/qubesvm.py", line 1596, in run_for_stdio
    raise subprocess.CalledProcessError(p.returncode,
subprocess.CalledProcessError: Command 'pw-play --format=f32 --rate=44100 --channels=1 audio_in.snd' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 309, in test_226_audio_playback_pipewire
    self.common_audio_playback()
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 151, in common_audio_playback
    self.fail('{} stderr: {}'.format(str(err), err.stderr))
AssertionError: Command 'pw-play --format=f32 --rate=44100 --channels=1 audio_in.snd' returned non-zero exit status 1. stderr: b'error: pw_context_connect() failed: Host is down\n'
```
2. Playback was somehow distorted on Fedora (but recording seems to work):
```
Traceback (most recent call last):
  File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 309, in test_226_audio_playback_pipewire
    self.common_audio_playback()
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 156, in common_audio_playback
    self.check_audio_sample(recorded_audio.file.read(), sfreq)
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 122, in check_audio_sample
    self.fail('frequency {} not in specified range'
AssertionError: frequency 2073.4488161685595 not in specified range
```
It's supposed to be 4400. I do not have actual recorded audio from the test to listen.

DemiMarie:
> https://openqa.qubes-os.org/tests/69017
> 
> I'm not sure if the issue is with tests, or the agent, but looks more likely to be an agent issue, so commenting here:
> 
> 1. Pipewire doesn't start on Debian:
> 
> ```
> Traceback (most recent call last):
>   File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 149, in common_audio_playback
>     self.loop.run_until_complete(self.testvm1.run_for_stdio(cmd))
>   File "/usr/lib64/python3.11/asyncio/base_events.py", line 653, in run_until_complete
>     return future.result()
>            ^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.11/site-packages/qubes/vm/qubesvm.py", line 1596, in run_for_stdio
>     raise subprocess.CalledProcessError(p.returncode,
> subprocess.CalledProcessError: Command 'pw-play --format=f32 --rate=44100 --channels=1 audio_in.snd' returned non-zero exit status 1.
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
>     return func(*args, **kwds)
>            ^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 309, in test_226_audio_playback_pipewire
>     self.common_audio_playback()
>   File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 151, in common_audio_playback
>     self.fail('{} stderr: {}'.format(str(err), err.stderr))
> AssertionError: Command 'pw-play --format=f32 --rate=44100 --channels=1 audio_in.snd' returned non-zero exit status 1. stderr: b'error: pw_context_connect() failed: Host is down\n'
> ```

[Debian 11 does not support using PipeWire for audio.][1]  The test should probably be skipped there.

> 2. Playback was somehow distorted on Fedora (but recording seems to work):
> 
> ```
> Traceback (most recent call last):
>   File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
>     return func(*args, **kwds)
>            ^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 309, in test_226_audio_playback_pipewire
>     self.common_audio_playback()
>   File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 156, in common_audio_playback
>     self.check_audio_sample(recorded_audio.file.read(), sfreq)
>   File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 122, in check_audio_sample
>     self.fail('frequency {} not in specified range'
> AssertionError: frequency 2073.4488161685595 not in specified range
> ```
> 
> It's supposed to be 4400. I do not have actual recorded audio from the test to listen.

If it works with PulseAudio emulation then this is probably a bug in the test.  I think I would notice if `pw-record - | pw-play -` played back my voice an octave lower :laughing:.

[1]: https://wiki.debian.org/PipeWire#Debian_11

marmarek:
> [Debian 11 does not support using PipeWire for audio.](https://wiki.debian.org/PipeWire#Debian_11) The test should probably be skipped there.

Ah, okay. I noticed pipewire-pulse package is not there, but pipewire itself is. Anyway, can you add skip in the test (just match the template name)? In this case, whonix-ws-16 too (as it's based on debian-11).

DemiMarie:
> This might be about when cork/uncork commands are sent and/or how they are handled by the daemon side.

I think this is exactly right.  If I do not send the cork command until the vchan has drained, the problem happens much less frequently.  If I do not send the cork command at all, the problem almost never happens.

pipewire/qubes-pw-module.c
```diff
@@ -497,17 +496,36 @@ static int process_control_commands(struct impl *impl)
...
+           if (!impl->cork_enable) {
+               playback_stream->last_state = new_state;
+               break;
+           }
+
+           if (playback_stream->vchan) {
+               int space = libvchan_buffer_space(playback_stream->vchan);
+               if (space >= 0 && (size_t)space != playback_stream->buffer_size) {
+                   spa_assert_se((size_t)space < playback_stream->buffer_size);
+                   pw_log_debug("will cork stream when vchan is drained (currently %zu bytes left)",
```
marmarek, line 512: How cork cmd will be sent? I can't see it retrying later.

marmarek:
> @DemiMarie
[Disable cork messages by default](https://github.com/QubesOS/qubes-gui-agent-linux/pull/157/commits/4626c818b9b3d95f583b203c7dcb64cd32875666)

This is not a good idea, the cork command was introduced for a reason: https://github.com/QubesOS/qubes-issues/issues/1549

pipewire/qubes-pw-module.c
DemiMarie: `process_control_commands()` is called whenever activity happens on either vchan, unless that vchan is closed.

marmarek:
I tested it on some other system and not corking at all is a bad idea - I hear the "silence" (some quiet noise) when the output is active but nothing is playing (probably a hardware bug, but still).

PR #158, opened 2022-06-07
Adjust pulseaudio script to retrieve pulsecore headers, adjust pulseaudio module references

ejose19:


ejose19:
Script to download pulsecore headers is adjusted from [this commit](https://github.com/QubesOS/qubes-gui-agent-linux/pull/71#issue-493727711).

Adjusted module location as mentioned [here](https://github.com/QubesOS/qubes-issues/issues/7531#issuecomment-1146800086) (due to pulseaudio 16.0 [changes](https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/16.0/#moduleinstallationlocationchangedremembertoupgradepaprefstothelatestversion))

I still see a reference to `pulse-$version` in `debian/pulseaudio-qubes.install` but I'm not sure how to proceed there.

fepitre:
There is already something here https://github.com/QubesOS/qubes-gui-agent-linux/blob/master/get-latest-pulsecore.sh

ejose19:
> There is already something here https://github.com/QubesOS/qubes-gui-agent-linux/blob/master/get-latest-pulsecore.sh

Wouldn't we better using upstream directly than proxying through fedora repos? I just checked [pulseaudio](https://packages.fedoraproject.org/pkgs/pulseaudio/pulseaudio/) version and there's no `16.0` yet, so we'd need to wait for fedora to list it there first.

fepitre:
> > There is already something here https://github.com/QubesOS/qubes-gui-agent-linux/blob/master/get-latest-pulsecore.sh
> 
> Wouldn't we be better using upstream directly than proxying through fedora repos? I just checked [pulseaudio](https://packages.fedoraproject.org/pkgs/pulseaudio/pulseaudio/) version and there's no `16.0` yet, so we'd need to wait for fedora to list it there first.

I was meaning there exist already similar script so we are going to add another one unless adapting the original one. Second, using Fedora was a way to verify and rely on Fedora content. So typically the script is downloading RPM, verifying the signature and then we process the content. In your approach there is not even signature check on the downloaded content so current form is not acceptable even if using upstream is a good idea.

ejose19:
> > > There is already something here https://github.com/QubesOS/qubes-gui-agent-linux/blob/master/get-latest-pulsecore.sh
> > 
> > 
> > Wouldn't we be better using upstream directly than proxying through fedora repos? I just checked [pulseaudio](https://packages.fedoraproject.org/pkgs/pulseaudio/pulseaudio/) version and there's no `16.0` yet, so we'd need to wait for fedora to list it there first.
> 
> I was meaning there exist already similar script so we are not going to add another one unless adapting the original one. Second, using Fedora was a way to verify and rely on Fedora content. So typically the script is downloading RPM, verifying the signature and then we process the content (yes we assume that packagers in distributions do their work to double check the sources). In your approach there is not even signature check on the downloaded content so current form is not acceptable even if using upstream is a good idea.

Yes I got your point, was just brainstorming how to get more recent sources, although if you require signature verification from a trusted authority the options are few.

I see `pulseaudio` tags are signed from

```
PulseAudio 16.0
gpg: Signature made Sat 28 May 2022 12:23:42 PM -03
gpg:                using RSA key 52DFA7B8BAC74687C8A88EF48165E3D1987E2132
gpg:                issuer "arun@asymptotic.io"
```
& for v10-13
```
PulseAudio 13.0
gpg: Signature made Fri 13 Sep 2019 10:20:58 AM -03
gpg:                using DSA key B61E1D411D57BD16F11536162477064CE8B9F3BD
gpg:                issuer "tanuk@iki.fi"
```

would adding these keys to the script and doing signature verification on the cloned repo be enough?

marmarek:
> would adding these keys to the script and doing signature verification on the cloned repo be enough? 

I think it's fine (TOFU verification approach).

> I'd just adjust the script you mentioned and remove the new one which is redundant.

Yes, that's ok.

ejose19:
Should we also make a CI script to run `get-latest-pulsecore.sh` on daily basis? The script would also update `archlinux/PKGBUILD` & `rpm_spec/gui-agent.spec.in` pulseaudio version (if applicable) as would be done in a manual update.

fepitre:
> Should we also make a CI script to run `get-latest-pulsecore.sh` on daily basis? The script would also update `archlinux/PKGBUILD` & `rpm_spec/gui-agent.spec.in` pulseaudio version (if applicable) as would be done in a manual update.

Right now it's my internal CI who take care of opening PR for kernel and pulseaudio at the same time, once a week. I can adapt that change it to a daily basis for pulseaudio.

ejose19:
> > Should we also make a CI script to run `get-latest-pulsecore.sh` on daily basis? The script would also update `archlinux/PKGBUILD` & `rpm_spec/gui-agent.spec.in` pulseaudio version (if applicable) as would be done in a manual update.
> 
> Right now it's my internal CI who take care of opening PR for kernel and pulseaudio at the same time, once a week. I can adapt that change it to a daily basis for pulseaudio.

Alright, hopefully with this new upstream source and the interval update, no more manual pulseaudio headers PRs will be needed (especially as archlinux updates so fast), yielding to more secure and less error-prone updates.

locriacyber:
Using Arch Linux. I rebuilt
`/usr/lib/pulse-15.0/modules/module-vchan-sink.so`
and moved it to
`/usr/lib/pulseaudio/modules/module-vchan-sink.so` (pulseaudio 16.0-1 on Arch)

However, I didn't see any modification in this PR, so I don't know if it will work.

get-latest-pulsecore.sh
```diff
@@ -21,28 +21,35 @@ if [ ! -d "$BUILDERDIR" ]; then
     exit 1
 fi
 
-if [ -z "$LATEST_FEDORA_RELEASE" ]; then
-    LATEST_FEDORA_RELEASE="$(git ls-remote --heads https://src.fedoraproject.org/rpms/fedora-release | grep -Po "refs/heads/f[0-9][1-9]*" | sed 's#refs/heads/f##g' | sort -g | tail -1)"
-fi
-LATEST_FEDORA_VERREL="$(dnf -q repoquery pulseaudio --disablerepo=* --enablerepo=fedora --enablerepo=updates --releasever="$LATEST_FEDORA_RELEASE" | grep -Po "[0-9][1-9]*\.*[0-9]*\.*[0-9]*-[0-9]*" | sort -V | tail -1)"
+REPO_URL=git://anongit.freedesktop.org/pulseaudio/pulseaudio
```
marmarek, line 24: Please https url.

get-latest-pulsecore.sh
```diff
@@ -21,28 +21,35 @@ if [ ! -d "$BUILDERDIR" ]; then
...
 
-LATEST_FEDORA_VERSION="$(echo "$LATEST_FEDORA_VERREL" | cut -d'-' -f1)"
+LATEST_REPO_VERSION="$(git ls-remote --exit-code --refs --tags --sort="v:refname" "$REPO_URL" '*.*' | tail -n1 | cut -d/ -f3 | sed 's/^v//')"
 LATEST_QUBES_VERSION="$(find "$LOCALDIR/pulse" -type d -name "pulsecore-*" | sed "s|$LOCALDIR/pulse/pulsecore-||" | sort -g | tail -1)"
 
-SRC_RPM="pulseaudio-${LATEST_FEDORA_VERREL}.fc${LATEST_FEDORA_RELEASE}.src.rpm"
-SRC_FILE="pulseaudio-${LATEST_FEDORA_VERSION}.tar.xz"
```
marmarek, line 33: `exit_updater()` uses this variable, so it needs to be updated too.

marmarek (changes requested):
Thanks @ejose19, besides two minor comments to the script, it looks good.

marmarek:
PipelineRetry

fepitre:
Can we separate this PR from the script updater? I would love to try the bot doing the PR.

get-latest-pulsecore.sh
ejose19: Looks like gitlab is the source of truth for other ones (anongit, cgit) so changed url to that one instead.

get-latest-pulsecore.sh
ejose19: Looks like this variable is redundant now, removed.

ejose19:
> Can we separate this PR from the script updater? I would love to try the bot doing the PR for the headers.

I've removed `pulse/pulsecore-16.0` folder from this PR so it can be added by the bot.

ejose19:
> Should we also make a CI script to run `get-latest-pulsecore.sh` on daily basis? The script would also update `archlinux/PKGBUILD` & `rpm_spec/gui-agent.spec.in` pulseaudio version (if applicable) as would be done in a manual update.

Rebased commits so history reflects current changes, also removed this manual change so the bot updates these files too

marmarek (approved):

PR #162, opened 2022-07-08
Respect FocusIn/Out `mode` in a hacky way

locriacyber:
Attempt to recreate different focus "mode" with Xlib API. Works good enough.

@DemiMarie

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -1787,56 +1787,72 @@ static void take_focus(Ghandles * g, XID winid)
 
 }
 
+static void handle_focus_helper(Ghandles * g, XID winid, struct msg_focus msg)
+{
+    struct genlist *l;
+    bool use_take_focus = false;
+    bool input_hint = false;
+    if ( (l=list_lookup(windows_list, winid)) && (l->data) )
+        input_hint = ((struct window_data*)l->data)->input_hint;
+    else {
+        fprintf(stderr, "WARNING handle_focus: Window 0x%x data not initialized", (int)winid);
+        input_hint = true;
+    }
+    
+    if (msg.type == FocusIn) {
+        if (msg.mode == NotifyNormal) {
+            XRaiseWindow(g->display, winid);
+            
+
+            if ( (l=list_lookup(windows_list, winid)) && (l->data) ) {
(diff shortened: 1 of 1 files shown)


locriacyber:
No. I haven't used this separated patch though.

This still gets confused when VM transfer focus among its windows.

locriacyber:
This is still buggy. When window is killed with keyboard shortcut, Mouse focus is grabbed, but never returned.

DemiMarie:
Does e0265cd988e14ae98a7bb8eb18ca74b86f9100cf fix it?

locriacyber:
Not entirely. I don't know much about how server-client tracks mapped windows. I only know about the X part.

The current issue is only that in AppVM 1, sometimes the focused window is destroyed without notice, then the focus is transfered to another window in AppVM 2 (because it was previously focused in dom0), but AppVM 2 don't think it has focus.

I don't know how to fix this other than being the X server itself. We should probably retire the current gui-proxy-thingy for a proper solution (custom X server).

DemiMarie:
> I don't know how to fix this other than being the X server itself, and use custom WM/X server that is meant to be nested (don't know if we need XFCE4 inside AppVM). We should probably retire the current gui-proxy-thingy for a proper solution (custom X server). That would get other benefits like supporting multiple input devices of the same kind.

@locriacyber: What about having a dummy window that is given focus whenever a window loses focus?
@marmarek: Am I correct that a custom X server is out of the question?

locriacyber:
> What about having a dummy window that is given focus whenever a window loses focus?
Users probably expect to have focus returned to last window they used (when a window is closed).

DemiMarie:
> > What about having a dummy window that is given focus whenever a window loses focus?
> 
> Users probably expect to have focus returned to last window they used (when a window is closed).

My thought was that this window would get a `FocusIn` event from the GUI daemon if it belonged to the same qube.  Otherwise, it is better for no window to have focus.

marmarek:
> @locriacyber: What about having a dummy window that is given focus whenever a window loses focus?

I think gui-agent already does something like this. Look for `g->stub_win` usage.

locriacyber:
Superseded by #169

PR #164, opened 2022-07-16
Use larger vchan buffers

DemiMarie:
This ensures that pacat-simple-vchan will not need to block, and
therefore prevents overruns if it discards data rather than blocking.


marmarek (approved):

DemiMarie:
CentOS CI failure is an unrelated networking problem on the CI runners.

PR #166, opened 2022-07-30
Harden get-latest-pulsecore.sh

DemiMarie:
This uses stricter input validation and key handling.  It also uses
Sequoia instead of GnuPG for key fetching.


get-latest-pulsecore.sh
```diff
@@ -22,29 +22,45 @@ fi
...
+        echo "$key:6:" | gpg --import-ownertrust
+        for i in keyserver.ubuntu.com keys.openpgp.org pgp.mit.edu keyserver.pgp.com; do
+            sq keyserver --server "$i" get --binary -- "0x$key" && break
+        done
+    done | gpg --homedir=../gnupg-tmp --import --no-armor
+    gpg --homedir=../gnupg-tmp --export -- "${trusted_signers[@]}" | gpg --import --no-armor
+
+    for key in "${trusted_signers[@]}"; do
+        echo "$key:6:"
+    done | gpg --import-ownertrust
```
marmarek, line 57: Why two keyrings and all this dance?
But also, I don't like using _both_ sequoia and gnupg. If sequoia can handle all the tasks (including git tag verification), then switch completely, otherwise stick with gnupg (and perhaps commit pubkeys into the repo to avoid keyservers interaction).

marmarek (changes requested):

get-latest-pulsecore.sh
marmarek: Besides the thing above (which is still relevant), why _two_ temporary gpg's homedirs? Does `sq keyserver get` potentially fetches a different key than it was asked for?

DemiMarie: I think I can use the chameleon version of Sequoia.

PR #169, opened 2022-08-11
XInput support for client

locriacyber:


locriacyber:
This crash for some reason
Logs:
[Xorg.0.log](https://github.com/QubesOS/qubes-gui-daemon/files/9309649/Xorg.0.log)
[.xsession-errors.txt](https://github.com/QubesOS/qubes-gui-daemon/files/9309624/default.xsession-errors.txt)

locriacyber:
@DemiMarie do you know what problem this is? It doesn't make much sense.

It has to do with `qubes-vm-gui` on arch linux, because it only happens after I installed it with this patch.

However, the code in the commit was copied from previous attempt, and the previous code worked fine.

DemiMarie:
@locriacyber I suggest `git bisect`.

PR #171, opened 2022-08-22
Give windows focus before sending events

DemiMarie:
Events from the GUI daemon are directed at specific windows.  Ensuring
that they are sent to the correct window is more importent than obeying
the ICCCM input hint.  Giving pointer focus to a window is accomplished
by raising it.


marmarek:
The "ignore input hint" part sounds suspicious. Support for it was added for a reason. See [this thread from 9 years ago](https://groups.google.com/g/qubes-devel/c/vGhdCsgQEbc/m/bTZZBlN7GAsJ) and verify if it works in that case too. Furthermore, some toolkits (GTK?) handle input events through a dedicated off-screen window instead of a window you see, I guess calling `XSetInputFocus` and ignoring their preferences will break them too. Worth testing with Firefox too (which I think uses this technique too).

DemiMarie:
What about merging QubesOS/qubes-gui-daemon#104?  That has been reported to help.

PR #173, opened 2022-10-05
Remove hanging person image

DemiMarie:
This could be offensive to some and it serves no purpose.

### DIFF
--- a/appvm-scripts/usrbin/qubes-session
+++ b/appvm-scripts/usrbin/qubes-session
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash --
 #
 # The Qubes OS Project, http://www.qubes-os.org
 #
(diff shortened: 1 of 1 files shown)


appvm-scripts/usrbin/qubes-session
```diff
@@ -75,11 +75,5 @@ fi
...
-# Now hang myself so Xorg may live :'(
-# ____
-# |/ |
-# |  O
-# | /|\
-# | / \
-# |
-env sleep inf
+# Sleep forever so that Xorg keeps running
+exec sleep inf
```
marmarek, line 79: Please keep it separate process (no `exec`), there may be some background children that could confuse sleep when exiting.

marmarek (changes requested):

marmarek (approved):

PR #174, opened 2022-10-21
Remove more checks for dispvm restore

DemiMarie:
Also switch from select(2) to poll(2).


marmarek:
Please split those two changes into separate commits. And fix build failure.

DemiMarie:
PipelineRetry

.gitignore
```diff
@@ -3,3 +3,11 @@ tags
...
 pkgs
 debian/changelog.*
+*.dep
+/gui-agent/qubes-gui
+/gui-common/qubes-gui-runuser
+*.orig
+*.rej
+*.gz
+*.zstd
+*.tgz
```
marmarek, line 13: Those aren't built products (or other files that are expected to be present and should be ignored), so don't add them.

Makefile
```diff
@@ -26,23 +26,42 @@ SYSLIBDIR ?= /lib
...
 
 appvm: gui-agent/qubes-gui gui-common/qubes-gui-runuser \
 	xf86-input-mfndev/src/.libs/qubes_drv.so \
 	xf86-video-dummy/src/.libs/dummyqbs_drv.so pulse/module-vchan-sink.so \
 	xf86-qubes-common/libxf86-qubes-common.so
 
-gui-agent/qubes-gui:
-	$(MAKE) -C gui-agent
+OBJS = vmside.o \
+	../gui-common/txrx-vchan.o ../gui-common/error.o ../common/list.o ../gui-common/encoding.o
```
marmarek, line 45: I don't like moving compilation of everything in `gui-agent/` into top level makefile. For example the above paths look misleading now, but also you pollute env for building other parts (xorg driver, pulseaudio etc) with agent-specific settings. You workaround some of it by using non-standard variable names, but that isn't a good idea.
But, since `gui-common/*` and `common/*`  is in fact used only in `gui-agent/` (not really shared with anything else), maybe simply move them all into `gui-agent/` dir?

gui-agent/vmside.c
```diff
@@ -2368,7 +2363,6 @@ int main(int argc, char **argv)
         } while (busy);
 
     }
-    return 0;
```
marmarek, line 2371: Please don't remove final return statement from non-void function, even if right now it's unreachable.

marmarek (changes requested):

Makefile
marmarek: This still applies.

gui-agent/txrx-vchan.c
```diff
@@ -84,47 +82,37 @@ int read_data(libvchan_t *vchan, char *buf, int size)
...
     if (ret < 0) {
-        perror("select");
+        if (errno == EINTR)
+            return -1;
+        perror("poll");
         exit(1);
     }
     if (!libvchan_is_open(vchan)) {
         fprintf(stderr, "libvchan_is_eof\n");
+        libvchan_close(vchan);
```
marmarek, line 97: You introduce double-free here.
Please don't combine subtle behavior changes with refactors (select->poll in this case).

gui-agent/txrx-vchan.c
```diff
@@ -84,47 +82,37 @@ int read_data(libvchan_t *vchan, char *buf, int size)
...
-    }
-    FD_SET(vfd, &rfds);
-    if (vfd > max)
-        max = vfd;
-    max++;
-    ret = select(max, &rfds, NULL, NULL, NULL);
-    if (ret < 0 && errno == EINTR)
-        return 0;
+    int ret;
+    write_data(vchan, NULL, 0);    // trigger write of queued data, if any present
```
marmarek, line 88: Another subtle behavior change with no even justification in commit message...

gui-agent/qubes-gui-runuser.c
```diff
@@ -320,5 +324,18 @@ int main(int argc, char **argv) {
     signal(SIGTERM, propagate_signal);
     signal(SIGHUP, propagate_signal);
 
-    return do_execute(argv[1], argv[2], argv+3);
+    char *user = argv[1];
+    unsigned int len = 0;
+    if (user[0] == 0) {
```
marmarek, line 329: This behavior needs to be documented somewhere. At the very least in the `usage()`.

marmarek (changes requested):

marmarek (approved):

PR #175, opened 2022-12-02
Add automatic clipboard wipe

mati7337:
Add functionality that automatically wipes the clipboard after a minute from the last vm paste.
https://github.com/QubesOS/qubes-issues/issues/3415

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -52,6 +52,9 @@
 #include "encoding.h"
 #include <libvchan.h>
 
+/* Time in milliseconds after which the clipboard data should be wiped */
+#define CLIPBOARD_WIPE_TIME 60000
+
 /* Get the size of an array.  Error out on pointers. */
 #define QUBES_ARRAY_SIZE(x) (0 * sizeof(struct { \
     int tried_to_compute_number_of_array_elements_in_a_pointer: \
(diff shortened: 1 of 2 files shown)


marmarek:
The thing you implemented is wiping the clipboard 1min after it get pasted to the VM or to an application (whichever comes later).  I think it's okay, but it needs to be documented (please open PR to qubes-doc repo, maybe to this page https://www.qubes-os.org/doc/how-to-copy-and-paste-text/ ?).

And also this should be configurable, at least enabling/disabling the feature. Since gui-agent doesn't have configuration file on its own, you can probably use qvm-service framework, which results in flag file present (or not) in `/run/qubes-service/`.

mati7337:
Should the clipboard wipe be by default enabled or disabled?

marmarek:
> Should the clipboard wipe be by default enabled or disabled?

I'd say default disabled.

gui-agent/vmside.c
```diff
@@ -2240,6 +2256,12 @@ int main(int argc, char **argv)
 
     parse_args(&g, argc, argv);
 
+    /* Clipboard wipe functionality is controlled by the
+     * qubes-gui-agent-clipboard-wipe service
+     */
+    g.clipboard_wipe =
+        access("/run/qubes-service/qubes-gui-agent-clipboard-wipe", F_OK) == 0;
```
marmarek, line 2263: It would be nice to announce support for this using https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-features.html#services. But then, there is a catch: qubesdb keys are limited to 63 characters, so `/features-request/supported-service.<service-name>` needs to fit there, and with the current name it does not. Maybe drop `qubes-` prefix?

BTW, announcing in this repo is to be added in `appvm-scripts/etc/qubes/post-install.d/20-qubes-guivm-gui-agent.sh`

marmarek (changes requested):

marmarek (approved):

PR #176, opened 2023-02-28
gui-agent: die when xorg fails to start

meithecatte:
See also QubesOS/qubes-issues#8060

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -1983,7 +1983,15 @@ static pid_t do_execute_xorg(
 static void *xorg_waitpid_thread(void *p) {
     Ghandles *g = p;
     int status;
-    waitpid(g->x_pid, &status, 0);
+    while (waitpid(g->x_pid, &status, 0) < 0) {
+        if (errno != EINTR) {
+            /* Looking at waitpid(2), everything other than EINTR would
+             * indicate a logic error; log this and give up. */
+            perror("waitpid");
+            abort();
+        }
+    }
+
     if (atomic_load(&terminating)) {
         exit(0);
     } else {
@@ -2297,8 +2305,9 @@ int main(int argc, char **argv)
     vchan_register_at_eof(handle_guid_disconnect);
     send_protocol_version(g.vchan);
     g.x_pid = get_xconf_and_run_x(&g);
-    if (pthread_create(&waitpid_thread, NULL, xorg_waitpid_thread, &g) < 0) {
+    if ((errno = pthread_create(&waitpid_thread, NULL, xorg_waitpid_thread, &g)) != 0) {
         perror("pthread_create");
+        kill(g.x_pid, SIGTERM);
         exit(1);
     }
     mkghandles(&g);
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -22,6 +22,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <stdatomic.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -29,6 +30,7 @@
 #include <stdlib.h>
 #include <signal.h>
 #include <fcntl.h>
+#include <pthread.h>
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <sys/stat.h>
@@ -144,6 +146,7 @@ struct genlist *windows_list;
 struct genlist *embeder_list;
 typedef struct _global_handles Ghandles;
 Ghandles *ghandles_for_vchan_reinitialize;
+atomic_bool terminating;
 
 #define SKIP_NONMANAGED_WINDOW do {                                    \
     if (!list_lookup(windows_list, window)) {                          \
@@ -1976,13 +1979,32 @@ static pid_t do_execute_xorg(
     }
 }
 
-static void terminate_and_cleanup_xorg(Ghandles *g) {
+/* this cannot be a SIGCHLD signal handler because of signal safety */
+static void *xorg_waitpid_thread(void *p) {
+    Ghandles *g = p;
     int status;
+    waitpid(g->x_pid, &status, 0);
+    if (atomic_load(&terminating)) {
+        exit(0);
+    } else {
+        if (WIFEXITED(status)) {
+            fprintf(stderr, "Xorg exited unexpectedly (exit code %d)\n",
+                    WEXITSTATUS(status));
+        } else if (WIFSIGNALED(status)) {
+            fprintf(stderr, "Xorg terminated unexpectedly (signal %d)\n",
+                    WTERMSIG(status));
+        } else {
+            fprintf(stderr, "Xorg terminated unexpectedly\n");
+        }
 
+        exit(1);
+    }
+}
+
+static void terminate_and_cleanup_xorg(Ghandles *g) {
     if (g->x_pid != (pid_t)-1) {
+        atomic_store(&terminating, true);
         kill(g->x_pid, SIGTERM);
-        waitpid(g->x_pid, &status, 0);
-        g->x_pid = -1;
     }
 }
 
@@ -2185,7 +2207,7 @@ static void handle_sigterm()
 {
     Ghandles *g = ghandles_for_vchan_reinitialize;
     terminate_and_cleanup_xorg(g);
-    exit(0);
+    while (1) pause();
 }
 
 static void usage()
@@ -2253,6 +2275,7 @@ int main(int argc, char **argv)
     int xfd;
     Ghandles g;
     int wait_fds[2];
+    pthread_t waitpid_thread;
 
     parse_args(&g, argc, argv);
 
@@ -2274,6 +2297,10 @@ int main(int argc, char **argv)
     vchan_register_at_eof(handle_guid_disconnect);
     send_protocol_version(g.vchan);
     g.x_pid = get_xconf_and_run_x(&g);
+    if (pthread_create(&waitpid_thread, NULL, xorg_waitpid_thread, &g) < 0) {
+        perror("pthread_create");
+        exit(1);
+    }
     mkghandles(&g);
     ghandles_for_vchan_reinitialize = &g;
     /* Turn on Composite for all children of root window. This way X server
@@ -2315,7 +2342,6 @@ int main(int argc, char **argv)
         fprintf(stderr, "XFixes not available, cursor shape handling off");
 
     XAutoRepeatOff(g.display);
-    signal(SIGCHLD, SIG_IGN);
     signal(SIGTERM, handle_sigterm);
     windows_list = list_new();
     embeder_list = list_new();


gui-agent/vmside.c
```diff
@@ -2188,6 +2188,12 @@ static void handle_sigterm()
     exit(0);
 }
 
+static void handle_sigchld()
+{
+    fprintf(stderr, "Xorg died unexpectedly, exiting!\n");
+    exit(1);
+}
+
```
DemiMarie, line 2196: `fprintf` and `exit` are not async-signal-safe, so one cannot call them from a signal handler that could interrupt another async-signal-unsafe function.  If one uses `sigprocmask` or `pthread_sigmask` to block SIGCHLD except during `ppoll()`, then the problem goes away, as `ppoll()` is async-signal-safe.  For this to work, SIGCHLD must be blocked before the handler is registered.

gui-agent/vmside.c
```diff
@@ -2188,6 +2188,12 @@ static void handle_sigterm()
     exit(0);
 }
 
+static void handle_sigchld()
```
DemiMarie, line 2191:
```suggestion
static void handle_sigchld(void)
```

Otherwise recent clang will (rightly) complain.  That said, one should really use the `siginfo_t` argument to check that it was in fact Xorg that exited, as opposed to e.g. a signal sent from another process.  Also, Xorg’s exit code should be logged.

gui-agent/vmside.c
```diff
@@ -2255,6 +2261,7 @@ int main(int argc, char **argv)
     int wait_fds[2];
 
     parse_args(&g, argc, argv);
+    signal(SIGCHLD, handle_sigchld);
```
DemiMarie, line 2264: Using `signal` to set a signal handler is deprecated.  Use `sigaction`.

DemiMarie (changes requested):
There are several bugs related to signal handling that need to be fixed.

marmarek:
There is also another issue: https://openqa.qubes-os.org/tests/67828#step/guivm_startup/21
```
Feb 28 20:05:22 sys-gui qubes-gui[476]: Xorg died unexpectedly, exiting!
```

I'm not yet sure if that just reported a crash elsewhere, or if that script is supposed to exit in case of sys-gui. I'll try to extract more logs from there.

marmarek:
This also makes stopping `qubes-gui-agent` always unclean (as it hits this signal handler when waiting for Xorg to terminate).

marmarek:
> I'm not yet sure if that just reported a crash elsewhere, or if that script is supposed to exit in case of sys-gui. I'll try to extract more logs from there.

Ok, this was stupid: https://github.com/QubesOS/qubes-gui-agent-linux/pull/177

gui-agent/vmside.c
meithecatte: I can't see the `ppoll` call you're talking about.  I do see one to `select`, though. Considering it's in `gui-common/txrx-vchan.c`, though, I'd gather that doing that there would be quite painful. Perhaps it'd be better to set a flag that the main loop would then check? Not sure whether we'd get stuck in `wait_for_vchan_or_argfd`, though. If we'll exit that function with a dead Xorg, I could check the flag in the main loop quite comfortably, but otherwise this seems like a painful thing.

gui-agent/vmside.c
meithecatte: Should we do anything about the use of `signal` for `SIGTERM`?

gui-agent/vmside.c
DemiMarie: You can use `pselect()` or the self-pipe trick.

gui-agent/vmside.c
DemiMarie: That should be fixed too.

DemiMarie (changes requested):

meithecatte:
Hmm, the approach of the signal handler just setting some kind of flag won't work, as in the motivating case, the signal gets delivered while the gui agent is stuck in `mkghandles`. And to be honest, I don't know how much I'd trust it to not get stuck in Xlib when Xorg dies, between the calls to `select` at which we could handle this.

I considered rewriting the signal handler for `SIGCHLD` to use signal-safe functions, but `snprintf` is not one of them, and using `_exit` isn't very nice either.

Another option would be to have the `SIGCHLD` handler attempt notifying us via a self-pipe, but also set an alarm in case that fails? And in that case, I guess we don't have to feel as bad about using `_exit`...

And then there's the option of starting up another thread, just in case Xorg dies on us, and handling this there?

I don't think there is an elegant solution here. Unix was a mistake.

DemiMarie:
> Hmm, the approach of the signal handler just setting some kind of flag won't work, as in the motivating case, the signal gets delivered while the gui agent is stuck in `mkghandles`. And to be honest, I don't know how much I'd trust it to not get stuck in Xlib when Xorg dies, between the calls to `select` at which we could handle this.

Xorg exiting should cause I/O to fail, so if Xlib hangs that is an Xlib bug.  Might be better to port the whole thing to XCB, which should at least be somewhat predictable.

> I considered rewriting the signal handler for `SIGCHLD` to use signal-safe functions, but `snprintf` is not one of them, and using `_exit` isn't very nice either.

It gets worse: due to race conditions, it isn’t safe to exit until all `MSG_WINDOW_DUMP` messages have been acknowledged.  So the event loop needs to keep running a bit longer.

> Another option would be to have the `SIGCHLD` handler attempt notifying us via a self-pipe, but also set an alarm in case that fails? And in that case, I guess we don't have to feel as bad about using `_exit`...

Self-pipe is safe, alarm isn’t.

> And then there's the option of starting up another thread, just in case Xorg dies on us, and handling this there?

Does the agent call `fork()` and then do anything that isn’t async-signal-safe before `execve()`?  If so, that will need to be dealt with first.

> I don't think there is an elegant solution here. Unix was a mistake.

Yes, it was.

meithecatte:
> Self-pipe is safe, alarm isn’t.

`alarm(2)` is listed in [`signal-safety(7)`](https://man7.org/linux/man-pages/man7/signal-safety.7.html), though?

> Does the agent call fork() and then do anything that isn’t async-signal-safe before execve()? If so, that will need to be dealt with first.

Could you expand on this? I don't really understand why.

> It gets worse: due to race conditions, it isn’t safe to exit until all `MSG_WINDOW_DUMP` messages have been acknowledged. So the event loop needs to keep running a bit longer.

What happens when we exit too early? I don't think we can rely on this either way, as this is happening across a security boundary.

> > I don't think there is an elegant solution here. Unix was a mistake.
>
> Yes, it was.

I'm glad we agree.

DemiMarie:
> > Self-pipe is safe, alarm isn’t.
> 
> `alarm(2)` is listed in [`signal-safety(7)`](https://man7.org/linux/man-pages/man7/signal-safety.7.html), though?

See below.  The tl;dr is that we _really_ do not want to exit uncleanly.

> > Does the agent call fork() and then do anything that isn’t async-signal-safe before execve()? If so, that will need to be dealt with first.
> 
> Could you expand on this? I don't really understand why.

`fork()` interacts badly with locks.  Any locks that were held in the parent process will remain held in the child process, but the threads that would unlock them do not exist in the child.  Since e.g. `malloc()` uses locks internally, this could lead to a deadlock in the child process.  Therefore, POSIX states that if the parent process is multi-threaded, the child process after `fork()` is only allowed to use async-signal-safe interfaces until `execve()`.  Under glibc, one can generally get away with `malloc()`, but it’s best not to rely on that.

> > It gets worse: due to race conditions, it isn’t safe to exit until all `MSG_WINDOW_DUMP` messages have been acknowledged. So the event loop needs to keep running a bit longer.
> 
> What happens when we exit too early? I don't think we can rely on this either way, as this is happening across a security boundary.

See [this GUI daemon commit](https://github.com/QubesOS/qubes-gui-daemon/pull/128/commits/a67185f91392852040fa248d531493c993932bd4), which will (hopefully) be merged eventually.  In short, while there is no security problem, it could cause a guest-wide hang or loss of network connectivity.  Ideally, the agent would go even further, and wait for all windows to be unmapped on the agent side. 

> > > I don't think there is an elegant solution here. Unix was a mistake.
> > 
> > Yes, it was.
> 
> I'm glad we agree.

Unix got some things right, but a lot of things wrong.

marmarek:
> Therefore, POSIX states that if the parent process is multi-threaded, the child process after `fork()` is only allowed to use async-signal-safe interfaces until `execve()`.

Citation needed.

Anyway, that's one of the thing such application must take care of. More complex application (which this isn't really) have APIs to register callbacks around fork (before, after in parent, after in child etc). Anyway, I don't think any of this applies here, because we aren't talking about interacting with things after fork() (but - if we would, xenstore may need some special care, as it may use threads).

> See [this GUI daemon commit](https://github.com/QubesOS/qubes-gui-daemon/pull/128/commits/a67185f91392852040fa248d531493c993932bd4), which will (hopefully) be merged eventually. In short, while there is no security problem, it could cause a guest-wide hang or loss of network connectivity. Ideally, the agent would go even further, and wait for all windows to be unmapped on the agent side.

But you do realize we are talking about handling premature Xorg exit here, right? At this point all grants are unmapped already. So, there is no point in complicating things in this case.

meithecatte:
> But you do realize we are talking about handling premature Xorg exit here, right? At this point all grants are unmapped already. So, there is no point in complicating things in this case.

Say Xorg just randomly segfaults. It's a pile of C code so it's not that farfetched. What part of the system unmaps the grants in this case?

DemiMarie:
> > See [this GUI daemon commit](https://github.com/QubesOS/qubes-gui-daemon/pull/128/commits/a67185f91392852040fa248d531493c993932bd4), which will (hopefully) be merged eventually. In short, while there is no security problem, it could cause a guest-wide hang or loss of network connectivity. Ideally, the agent would go even further, and wait for all windows to be unmapped on the agent side.
> 
> But you do realize we are talking about handling premature Xorg exit here, right? At this point all grants are unmapped already. So, there is no point in complicating things in this case.

Whoops!  I forgot that the grants are mapped by Xorg, not the agent process.

marmarek:
> What part of the system unmaps the grants in this case?

Kernel, on FD release.

meithecatte:
I pushed a much larger diff, that should handle all this properly.

gui-agent/vmside.c
```diff
@@ -1976,13 +1979,32 @@ static pid_t do_execute_xorg(
     }
 }
 
-static void terminate_and_cleanup_xorg(Ghandles *g) {
+/* this cannot be a SIGCHLD signal handler because of signal safety */
+static void *xorg_waitpid_thread(void *p) {
+    Ghandles *g = p;
     int status;
+    waitpid(g->x_pid, &status, 0);
```
DemiMarie, line 1986: What if this fails?

gui-agent/vmside.c
```diff
@@ -1976,13 +1979,32 @@ static pid_t do_execute_xorg(
...
+    if (atomic_load(&terminating)) {
+        exit(0);
```
DemiMarie, line 1988: Should the exit status really be zero here?

DemiMarie (commented):
The check for the return code of

gui-agent/vmside.c
meithecatte: This happens if we get SIGTERM. Not sure if exit(0) is what we should do in this case, but it is what the previous code did. What would you suggest?

DemiMarie: Not sure tbh.

meithecatte: Maybe we would want to reset the SIGTERM sigaction and send SIGTERM to ourselves, so that the parent gets told we were killed by SIGTERM? Seems like unnecessary complexity though, if I'm being honest.

gui-agent/vmside.c
meithecatte: Added some error handling here. Not sure if it's the *best* strategy, but at least we're not failing silently.

gui-agent/vmside.c
```diff
@@ -2274,6 +2297,10 @@ int main(int argc, char **argv)
     vchan_register_at_eof(handle_guid_disconnect);
     send_protocol_version(g.vchan);
     g.x_pid = get_xconf_and_run_x(&g);
+    if (pthread_create(&waitpid_thread, NULL, xorg_waitpid_thread, &g) < 0) {
+        perror("pthread_create");
+        exit(1);
+    }
```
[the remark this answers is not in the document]
meithecatte: Good point.

marmarek:
Uhm, I blinked and suddenly gui-agent got another thread. I must admit I don't really like it... I'd much more prefer the approach with a flag and checking it where necessary. In case of waiting for Xorg startup, I guess that would be mostly `wait_for_unix_socket()` (currently it will exit(1) if accept fails, including EINTR case).

marmarek:
Done with much simpler approach in https://github.com/QubesOS/qubes-gui-agent-linux/pull/219

PR #181, opened 2023-04-24
Support alternate user names

DemiMarie:
The default user might not be named `user`.

### DIFF
--- a/appvm-scripts/etc/securitylimits.d/90-qubes-gui.conf
+++ b/appvm-scripts/etc/securitylimits.d/90-qubes-gui.conf
@@ -1,4 +1,5 @@
-# Qubes GUI agent needs to mlock() all the composition buffers, so they are
-# constant physical address
+# Old Qubes GUI agent needs to mlock() all the composition buffers, so they are
+# constant physical addresses.  Current GUI agent needs to allocate large
+# amounts of pinned memory for grant sharing.
 user   soft memlock 131072
 user   hard memlock 131072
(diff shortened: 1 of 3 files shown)


marmarek (approved):

marmarek:
PipelineRetry

appvm-scripts/etc/securitylimits.d/90-qubes-gui.conf
```diff
@@ -1,4 +1,8 @@
...
-user   soft memlock 131072
-user   hard memlock 131072
+# Old Qubes GUI agent needs to mlock() all the composition buffers, so they are
+# constant physical addresses.  Current GUI agent needs to allocate large
+# amounts of pinned memory for grant sharing.
+@qubes   - memlock 131072
+
+# Qubes group is trusted and this makes audio handling easier
+@qubes   - nice -20
+@qubes   - rtprio unlimited
```
marmarek, line 8: Is it actually used by anything? Not sure about pipewire, but pulseaudio asks rtkit-daemon to change its priority, so this setting is not needed for it.

marmarek (changes requested):

appvm-scripts/etc/securitylimits.d/90-qubes-gui.conf
DemiMarie: I was having some sort of problem with rtkit, but this should be a separate PR.

marmarek: So, can you move this into separate PR, or drop completely until further diagnosed?

marmarek: ping?

marmarek (approved):

PR #183, opened 2023-04-26
Add pulse 15.99.1 for jammy build

unman:
Also needed in release4.1
Closes QubesOS/qubes-issues#7537


marmarek:
Please don't include *.c files, only headers. There is a script for that in the repository root (requires slight modification for getting explicit version, not the latest one). I'll fetch the headers in a moment.

PR #187, opened 2023-05-04
pipewire-qubes: add pipewire-pulse dependency

fepitre:
https://github.com/QubesOS/qubes-issues/issues/8160

### DIFF
--- a/.qubesbuilder
+++ b/.qubesbuilder
@@ -10,6 +10,9 @@ vm:
   deb:
     build:
     - debian
+    source:
+      commands:
+      - 'sed -i "s/, pulseaudio-qubes/, pipewire-qubes/" @SOURCE_DIR@/debian/control'
   archlinux:
     build:
     - archlinux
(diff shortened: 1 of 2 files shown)


marmarek:
@DemiMarie is this right thing to do? It means you can no longer install pipewire-qubes and pulseaudio-qubes at the same time, but maybe that's okay?

fepitre:
In all the cases, there is something to solve. We need to ensure that both pulseaudio and pipewire are not installed together with default presets. My previous issue with Ubuntu was that pulseaudio-qubes and pulseaudio works only when pipewire was stopped. Originally Ubuntu installed it by default and it is enabled too.

DemiMarie:
> @DemiMarie is this right thing to do? It means you can no longer install pipewire-qubes and pulseaudio-qubes at the same time, but maybe that's okay?

In Fedora, I would prefer to make PipeWire the default to match what upstream does.  This includes both dom0 and Fedora templates.  In that case I don’t think there is a problem with such a conflict, as nobody will be running PulseAudio in Fedora.

> My previous issue with Ubuntu was that pulseaudio-qubes and pulseaudio works only when pipewire was stopped. Originally Ubuntu installed it by default and it is enabled too.

Having both agents running indeed makes no sense.  Some form of mutual exclusion could be used.

marmarek (approved):

PR #193, opened 2023-08-10
Wayland support in `qubes-gui-runuser`

DemiMarie:
This adds Wayland support to `qubes-gui-runuser`.  X11 support is not affected, so the same launcher (with different arguments) can be used for both Wayland and X11 sessions.


marmarek:
This has now a conflict.

marmarek:
nvm, it was a trivial one

marmarek (approved):

PR #194, opened 2023-08-26
Create virtual input device to allow the use of kloak

skyzzuu:
Partial resolution of issue https://github.com/QubesOS/qubes-issues/issues/1850. Keyboard events are sent to the virtual input device, but mouse events are not yet.

This creates an input device with the name "Qubes Virtual Input Device" on gui agent startup and translates KeyPress events into EV_KEY events and writes them to /dev/uinput. This allows the use of kloak within a qube.

Mouse events are currently not translated or written to /dev/uinput yet as I'm having trouble getting the cursor in the right location, they are still handled by the feed_xdriver function.


skyzzuu:
When the input device is created, notifications are generated mentioning that qubes.InputKeyboard and qubes.InputMouse were denied originating from the qube that started. The input device still works fine.

This pull request in qubes-app-linux-input-proxy repository stops qubes-input-trigger from running on the virtual input device created, preventing the notifications from being triggered.

https://github.com/QubesOS/qubes-app-linux-input-proxy/pull/30

marmarek:
> Mouse events are currently not translated or written to /dev/uinput yet as I'm having trouble getting the cursor in the right location, they are still handled by the feed_xdriver function.

You may want to use a device with absolute events (`EV_ABS`) instead of relative ones. Otherwise it will be quite easy to get out of sync. But if going with relative events anyway, maybe query the current mouse position and move based on that? (be careful about race conditions then: some movement being already queued to the device, but not handled by Xorg yet).

gui-agent/vmside.c
```diff
@@ -1626,62 +1658,71 @@ static void handle_keypress(Ghandles * g, XID UNUSED(winid))
...
+                XFreeModifiermap(modmap);
             }
-            XFreeModifiermap(modmap);
         }
-    }
 
-    feed_xdriver(g, 'K', key.keycode, key.type == KeyPress ? 1 : 0);
+        feed_xdriver(g, 'K', key.keycode, key.type == KeyPress ? 1 : 0);
+    } else {
+        struct input_event iev;
```
marmarek, line 1720: Not syncing modifiers state here looks suspicious. What happens for example if you Alt-tab into another's VM window, and switch back but using mouse? Won't Alt remain stuck? There may be also other similar cases, like opening a menu with Alt-F1 etc, switching workspaces with some key pressed etc.

gui-agent/vmside.c
```diff
@@ -2279,6 +2320,71 @@ int main(int argc, char **argv)
     int xfd;
     Ghandles g;
     int wait_fds[2];
+    
+        
+    // open uinput, if it fails, falls back to xdriver to not break input to the qube
+    g.output_fd = open("/dev/uinput", O_WRONLY | O_NDELAY);
```
marmarek, line 2326: Please make this feature opt-in, not the default. You can use qvm-service framework for that - see how clipboard wipe is handled for example.

marmarek (changes requested):

skyzzuu:
I'll get those changes implemented. One thing I forgot is that I had to add user to the input group so that xinit had enough permissions to be able to access the device. I'll find the correct repository to implement that and send a pull for that as well. This pull request will be dependent on that request as well once it's in.

gui-agent/vmside.c
skyzzuu: Would it be alright for it to be on by default in whonix workstation qubes, and then opt-in for all other qubes? If not, it's no problem and I can make it opt-in for all.

marmarek:
> I'll get those changes implemented. One thing I forgot is that I had to add user to the input group so that xinit had enough permissions to be able to access the device. I'll find the correct repository to implement that and send a pull for that as well. This pull request will be dependent on that request as well once it's in.

It's probably just about adding appropriate udev rule, it can be done in this repo.

gui-agent/vmside.c
marmarek: There are ways to enable "qvm-service" in whonix by default, either via https://github.com/QubesOS/qubes-core-admin-addon-whonix/, or by simply touching relevant file in `ExecStartPre=` added with a service drop-in file.

skyzzuu: Sorry didn't realize this needed to be marked as resolved. I made it opt-in using a qvm-service with the name gui-agent-virtual-input-device in commit c2f6757c65247c63c8e1e5889dc324027bee0f2c

gui-agent/vmside.c
skyzzuu: I changed it so that the section in handle_keypress that syncs modifiers will run whether input device creation was requested or not in e6d5e592d04d33361eb48eaa5d314bb17c5d313f

marmarek (changes requested):

gui-agent/vmside.c
```diff
@@ -1680,8 +1713,16 @@ static void handle_keypress(Ghandles * g, XID UNUSED(winid))
             XFreeModifiermap(modmap);
         }
     }
-
-    feed_xdriver(g, 'K', key.keycode, key.type == KeyPress ? 1 : 0);
+        
+    if(!g->createdInputDevice) {
```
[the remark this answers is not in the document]
skyzzuu: I adjusted it now so when virtual input device creation has been requested, modifiers are synced through the uinput fd instead of the xserver fd.

They are completely separate from each other now:

if(virtual input device NOT requested){
// original handle_keypress
} else {
// new handle_keypress (only writes to the uinput fd, never to the xserver fd)
}


Before these changes, all of the modifiers worked correctly when switching between qubes, except when pressing down the modifier and switching into a qube.


Now, all of the following scenarios work as expected:

Start in qube 1
Hold down shift
Switch to qube 2
Start typing while still holding down shift
This results in uppercase characters being typed out into qube 2 as expected


Start in qube 1
Hold down shift
Switch to qube 2
Release shift
Switch back to qube 1
Start typing
This results in lowercase characters being typed out into qube 1 as expected

Open a text file in nano in qube 1
Switch to qube 2
Hold down ctrl
Switch back to qube 1
Press s while still holding down ctrl
nano receives a ctrl+s and attempts to save the file


Modifiers still work exactly as expected when not switching between qubes as well.

skyzzuu:
I adjusted it now so when virtual input device creation has been requested, modifiers are synced through the uinput fd instead of the xserver fd.

They are completely separate from each other now:

if(virtual input device NOT requested){
// original handle_keypress
} else {
// new handle_keypress (only writes to the uinput fd, never to the xserver fd)
}


Before these changes, all of the modifiers worked correctly when switching between qubes, except when pressing down the modifier and switching into a qube.


Now, all of the following scenarios work as expected:

Start in qube 1
Hold down shift
Switch to qube 2
Start typing while still holding down shift
This results in uppercase characters being typed out into qube 2 as expected


Start in qube 1
Hold down shift
Switch to qube 2
Release shift
Switch back to qube 1
Start typing
This results in lowercase characters being typed out into qube 1 as expected

Open a text file in nano in qube 1
Switch to qube 2
Hold down ctrl
Switch back to qube 1
Press s while still holding down ctrl
nano receives a ctrl+s and attempts to save the file


Modifiers still work exactly as expected when not switching between qubes as well.

appvm-scripts/lib/udev/rules.d/90-qubes-virtual-input-device.rules
```diff
@@ -0,0 +1 @@
+ATTRS{name}=="Qubes Virtual Input Device", ACTION=="add", GROUP="user"
```
marmarek, line 1: This file is not installed anywhere (add it to makefile, and then both rpm spec file and debian "install" file).
And it should be `GROUP="qubes"`

gui-agent/vmside.c
```diff
@@ -285,6 +295,29 @@ static int compare_supported_cursors(const void *a,
                   ((struct supported_cursor *)b)->name);
 }
 
+
+void send_event(Ghandles * g, const struct input_event *iev) {
```
marmarek, line 299:
```suggestion
static void send_event(Ghandles * g, const struct input_event *iev) {
```

gui-agent/vmside.c
```diff
@@ -2279,6 +2379,66 @@ int main(int argc, char **argv)
...
+            g.created_input_device = 0;
+        }
+        
+
+        
+        
+
+        // set all keys
+        for(int i = 1; i < KEY_MAX; i++) {
+            if(i != KEY_RESERVED && ioctl(g.uinput_fd, UI_SET_KEYBIT, i) < 0) {
```
marmarek, line 2417: With all keys set, udev (and so Xorg) detects this as a combo keyboard+mouse device now. See `udevadm info -q all -n /dev/input/event1` (or whatever device node it will get). It should not have `ID_INPUT_MOUSE` set. I guess you need to exclude `BTN_*` range (`BTN_MISC` .. `BTN_GEAR_UP` - various buttons of mouses, tablets, game pads etc)

marmarek (changes requested):
Functionally it looks mostly fine now (with remarks below). But last commits introduced several repeated blank lines and whitespaces at the end of empty lines. There is also small merge conflict now, please rebase on the updated main branch.

marmarek:
You still need to rebase to resolve the conflict.

skyzzuu:
It should be all set now.

marmarek:
I did some more tests and Caps Lock behaves funny. When it's enabled, only some letter are capital, not all. But disabling it when the VM has focus doesn't seems to work (then, all letters typed into VM are capital). I guess it's related to modifiers sync, as this one is press-and-release to change the state, not press-and-hold.

skyzzuu:
Got it working now. The line to flip its associated bit in last_known_modifier_states was missing and caps lock needed to be excluded from key press/release below.

gui-agent/vmside.c
```diff
@@ -1615,62 +1648,124 @@ static void handle_keypress(Ghandles * g, XID UNUSED(winid))
...
+        
+        feed_xdriver(g, 'K', key.keycode, key.type == KeyPress ? 1 : 0);
+    } else {
+        int mod_mask;
+        int mod_index;
+        struct input_event iev;
+        iev.type = EV_KEY;
+        // only used in here to see if modifier is disabled
+        XModifierKeymap *modmap;
+        modmap = XGetModifierMapping(g->display);
```
marmarek, line 1716: You need to handle also when this call fails (see above).

gui-agent/vmside.c
```diff
@@ -1615,62 +1648,124 @@ static void handle_keypress(Ghandles * g, XID UNUSED(winid))
...
+                    // send modifier press
+                    send_event(g, &iev);
+                    // update state for this modifier
+                    g->last_known_modifier_states ^= mod_mask;
                 }
             }
+        }
+        
+        if(key.keycode-8 != KEY_CAPSLOCK) {
             XFreeModifiermap(modmap);
```
marmarek, line 1762: freeing should be outside of this if, no?

marmarek (changes requested):

marmarek (approved):

PR #197, opened 2023-10-01
Allow hexadecimal numbers

DemiMarie:
This may be more correct than using the “paused” event.


marmarek:
Is this still relevant? If so, needs rebasing.

DemiMarie:
> Is this still relevant? If so, needs rebasing.

Yes, but with a completely different reason than the original PR.  Want me to make a whole new PR?

marmarek:
PipelineRetry

PR #198, opened 2023-10-03
Pipewire debug

marmarek:


DemiMarie (approved):
I see no reason to not go ahead and merge this.

PR #200, opened 2023-10-28
pipewire: adjust buffer parameters

marmarek:
Allow more than one buffer (to allow application / daemon preparing next
buffer when previous one is being processed). And also adjust buffer
size based on other pipewire modules.

QubesOS/qubes-issues#8415

### DIFF
--- a/pipewire/qubes-pw-module.c
+++ b/pipewire/qubes-pw-module.c
@@ -790,9 +790,9 @@ static void stream_param_changed(void *data, uint32_t id,
 doit:
     params[0] = spa_pod_builder_add_object(&b,
             SPA_TYPE_OBJECT_ParamBuffers, SPA_PARAM_Buffers,
-            SPA_PARAM_BUFFERS_buffers, SPA_POD_Int(1),
+            SPA_PARAM_BUFFERS_buffers, SPA_POD_CHOICE_RANGE_Int(8, 2, 64),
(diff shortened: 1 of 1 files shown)


pipewire/qubes-pw-module.c
```diff
@@ -790,9 +790,9 @@ static void stream_param_changed(void *data, uint32_t id,
 doit:
     params[0] = spa_pod_builder_add_object(&b,
             SPA_TYPE_OBJECT_ParamBuffers, SPA_PARAM_Buffers,
-            SPA_PARAM_BUFFERS_buffers, SPA_POD_Int(1),
+            SPA_PARAM_BUFFERS_buffers, SPA_POD_CHOICE_RANGE_Int(8, 2, 64),
             SPA_PARAM_BUFFERS_blocks, SPA_POD_Int(1),
-            SPA_PARAM_BUFFERS_size, SPA_POD_Int(2048),
+            SPA_PARAM_BUFFERS_size, SPA_POD_Int(2048 * (int)impl->frame_size),
```
DemiMarie, line 795: Is there a reason to worry about this overflowing?  I presume not.

DemiMarie (approved):
Looks good to me.

pipewire/qubes-pw-module.c
marmarek: Not really, frame_size is always 4 with the current format (and very small number with any other format, if we'd change).

PR #210, opened 2024-03-05
Allow carriage return in secure clipboard

DemiMarie:
Fixes copy & paste of data containing it, such as from Windows.

Fixes: QubesOS/qubes-issues#8937

### DIFF
--- a/gui-agent/encoding.c
+++ b/gui-agent/encoding.c
@@ -114,7 +114,8 @@ bool is_valid_clipboard_string_from_vm(unsigned char *untrusted_s)
     for (; *untrusted_s; untrusted_s++) {
         // allow only non-control ASCII chars
         if ((*untrusted_s >= 0x20 && *untrusted_s <= 0x7E) ||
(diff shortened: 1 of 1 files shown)


gui-agent/encoding.c
```diff
@@ -114,7 +114,8 @@ bool is_valid_clipboard_string_from_vm(unsigned char *untrusted_s)
     for (; *untrusted_s; untrusted_s++) {
         // allow only non-control ASCII chars
         if ((*untrusted_s >= 0x20 && *untrusted_s <= 0x7E) ||
-             *untrusted_s == '\n' || *untrusted_s == '\t')
+             *untrusted_s == '\n' || *untrusted_s == '\t' ||
+             *untrusted_s == '\r')
```
marmarek, line 118: I'd prefer to allow it only when accompanied by `\n`.

marmarek (changes requested):

marmarek (approved):

PR #211, opened 2024-03-26
Watch qdb entry being set and modified on start or while playing

fepitre:
https://github.com/QubesOS/qubes-issues/issues/8975

### DIFF
--- a/pipewire/qubes-pw-module.c
+++ b/pipewire/qubes-pw-module.c
@@ -147,6 +147,31 @@ PW_LOG_TOPIC_STATIC(mod_topic, "mod." NAME);
                         "[ stream.sink.props=<properties> ] " \
                         "[ stream.source.props=<properties> ] "
 
+// FIXME: this should be a Qubes-wide domID parsing function
+static int parse_number(const char *const str,
+        size_t max_value, size_t *res, const char *const msg)
+{
+    char *endptr = (void *)1;
+    errno = *res = 0;
+    unsigned long long value = strtoull(str, &endptr, 0);
+    if (errno) {
+        int i = errno;
+        pw_log_error("Invalid %s \"%s\": %m", msg, str);
+        return -i;
+    } else if (*endptr) {
+        pw_log_error("Invalid %s \"%s\": trailing junk (\"%s\")",
+                msg, str, endptr);
+        return -EINVAL;
+    } else if (value > max_value) {
+        pw_log_error("Invalid %s \"%s\": exceeds maximum %s %zu",
+                msg, str, msg, max_value);
+        return -ERANGE;
+    } else {
+        *res = (size_t)value;
+        return 0;
+    }
+}
+
 static const struct spa_dict_item module_props[] = {
     { PW_KEY_MODULE_AUTHOR, "Wim Taymans <wim.taymans@gmail.com>, "
                             "Demi Marie Obenour <demi@invisiblethingslab.com>" },
@@ -294,7 +319,11 @@ struct impl {
 
     struct qubes_stream stream[2];
 
-    uint32_t frame_size, domid;
+    qdb_handle_t qdb;
+    struct spa_source qdb_watch_source;
+
+    uint32_t frame_size;
+    int domid;
 
     bool do_disconnect;
 #if !QUBES_PW_HAS_SCHEDULE_DESTROY
@@ -303,7 +332,47 @@ struct impl {
 #endif
 
     atomic_uint_fast64_t reference_count;
+
+    const char *args;
 };
+
+int get_domid_from_props(struct pw_properties *props) {
+    size_t domid;
+    const char *peer_domain_prop = NULL;
+    if ((peer_domain_prop = pw_properties_get(props, QUBES_AUDIOVM_PW_KEY)) == NULL) {
+        return -EINVAL;
+    }
+    if (parse_number(peer_domain_prop, INT_MAX / 2, &domid, "domain ID")) {
+        pw_log_debug("Cannot parse domid");
+        return -errno;
+    }
+    return domid;
+}
+
+int get_domid_from_qdb(qdb_handle_t qdb) {
+    size_t domid;
+    int res;
+    char *qdb_entry = qdb_read(qdb, QUBES_AUDIOVM_QUBESDB_ENTRY, NULL);
+
+    if (qdb_entry != NULL) {
+        if (parse_number(qdb_entry, INT_MAX / 2, &domid, "domain ID")) {
+            pw_log_error("Cannot parse domid");
+            res = -errno;
+        } else {
+            res = (int)domid;
+        }
+    } else {
+        res = -errno;
+        if (res == -ENOENT)
+            pw_log_error("no %s entry in QubesDB", QUBES_AUDIOVM_QUBESDB_ENTRY);
+        else
+            pw_log_error("unable to obtain %s entry from QubesDB", QUBES_AUDIOVM_QUBESDB_ENTRY);
+    }
+
+    free(qdb_entry);
+
+    return res;
+}
 
 static inline void
 impl_incref(struct impl *impl)
@@ -356,6 +425,7 @@ static void stop_watching_vchan(struct qubes_stream *stream)
         stream->vchan = NULL;
     }
     stream->is_open = false;
+    stream->source.fd = -1;
     // Update the main-thread state asynchronously, but only if the stream
     // is not being torn down.
     if (!stream->dead) {
@@ -566,8 +636,13 @@ static int add_stream(struct spa_loop *loop,
 static int connect_stream(struct qubes_stream *stream)
 {
     const char *msg = qubes_stream_is_capture(stream) ? "capture" : "playback";
-    uint16_t domid = stream->impl->domid;
+    int32_t domid = (int32_t)stream->impl->domid;
     int status;
+    if (domid<0) {
+        pw_log_warn("unknown peer domain, cannot create stream");
+        return 0;
+    }
+    pw_log_info("module %p: new (%s), peer id is %" PRIi32, stream->impl, stream->impl->args, domid);
 
     spa_assert_se(stream->vchan == NULL);
     spa_assert_se(stream->closed_vchan == NULL);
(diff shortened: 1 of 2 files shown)


fepitre:
It might deserves more logging messages but this is a first draft approach. In order to have a smooth switching experience, it requires that (see https://github.com/QubesOS/qubes-gui-daemon/pull/141) `pacat-simple-vchan` remains open for welcoming any qube. For example, in my setup, I have both `dom0` and `sys-usb` having it running for my test qube. In that case, playing VLC switches perfectly fine between both when changing `audiovm`.

marmarek:
@DemiMarie

pipewire/qubes-pw-module.c
```diff
@@ -187,7 +212,7 @@ struct qubes_stream {
     // Stream event listener.  Accessed on main thread only.
     struct spa_hook stream_listener;
     // Position control.  Pointer is immutable after being set.
-    // Pointee accessed on realtime thread only.
+    // Pointer accessed on realtime thread only.
```
DemiMarie, line 215: I used “pointee” to refer to what the pointer points to, as opposed to the pointer itself.

pipewire/qubes-pw-module.c
```diff
@@ -211,6 +236,7 @@ struct qubes_stream {
     uint64_t next_time;
     struct spa_source timer;
     struct rate_match rm;
+
```
DemiMarie, line 239:
```suggestion
```

pipewire/qubes-pw-module.c
```diff
@@ -566,8 +628,25 @@ static int add_stream(struct spa_loop *loop,
...
+        pw_log_debug("using domid from pw props");
+        domid = domid_from_props;
+    } else if (domid_from_qdb>=0) {
+        pw_log_debug("using domid from qdb");
+        domid = domid_from_qdb;
+    }
+    
+    if (domid<0) {
+        pw_log_debug("no domid, cannot create stream");
+        return 0;
```
DemiMarie, line 646: This should be a negative number.

DemiMarie (commented):
Looks mostly good so far, but I’ll want to check tomorrow before I formally approve this change.

pipewire/qubes-pw-module.c
fepitre: At first, I thought returning negative value would exit main loop, but it looks like it is not the case. I've adjusted.

pipewire/qubes-pw-module.c
fepitre: Thank you, remaining typo.

pipewire/qubes-pw-module.c
fepitre: Corrected, to be honest, I thought it was a typo without actually reading it as I did it at the very beginning of this work without understanding yet the whole context.

pipewire/qubes-pw-module.c
```diff
@@ -303,8 +332,40 @@ struct impl {
...
+    return domid;
+}
+
+int get_domid_from_qdb(qdb_handle_t qdb) {
+    size_t domid;
+
+    char *qdb_entry = qdb_read(qdb, QUBES_AUDIOVM_QUBESDB_ENTRY, NULL);
+    if (qdb_entry != NULL) {
+        if (parse_number(qdb_entry, INT_MAX / 2, &domid, "domain ID")) {
+            pw_log_debug("Cannot parse domid");
```
DemiMarie, line 358: This should have `error` severity, as it means something is quite wrong.  Please log the entry, too.

pipewire/qubes-pw-module.c
```diff
@@ -835,6 +913,65 @@ static void vchan_ready(struct spa_source *source)
...
+
+    pw_log_debug("Received event from QubesDB");
+
+    domid = get_domid_from_qdb(impl->qdb);
+
+    if (domid != impl->domid) {
+        for (int i = 0; i < 2; ++i) {
+            if (impl->stream[i].source.fd != -1)
+                pw_log_error("closing %d", impl->stream[i].source.fd);
+            spa_loop_invoke(impl->data_loop, vchan_reconnect_cb, 0, NULL, 0, true, &impl->stream[i]);
```
DemiMarie, line 941: If the return value is `-EPIPE`, this needs to be retried.

pipewire/qubes-pw-module.c
```diff
@@ -303,8 +332,40 @@ struct impl {
...
+    size_t domid;
+    const char *peer_domain_prop = NULL;
+    if ((peer_domain_prop = pw_properties_get(props, QUBES_AUDIOVM_PW_KEY)) == NULL) {
+        return -EINVAL;
+    }
+    if (parse_number(peer_domain_prop, INT_MAX / 2, &domid, "domain ID")) {
+        pw_log_debug("Cannot parse domid");
+        return -errno;
+    }
+    return domid;
```
DemiMarie, line 349: Please document in the commit message that `QUBES_AUDIOVM_PW_KEY` is no longer set.

pipewire/qubes-pw-module.c
```diff
@@ -303,8 +332,40 @@ struct impl {
...
+}
+
+int get_domid_from_qdb(qdb_handle_t qdb) {
+    size_t domid;
+
+    char *qdb_entry = qdb_read(qdb, QUBES_AUDIOVM_QUBESDB_ENTRY, NULL);
+    if (qdb_entry != NULL) {
+        if (parse_number(qdb_entry, INT_MAX / 2, &domid, "domain ID")) {
+            pw_log_debug("Cannot parse domid");
+            return -errno;
```
DemiMarie, line 359: Leaks `qdb_entry`.

DemiMarie (changes requested):
Thanks for the PR!  I hope I was not too harsh reviewing.

pipewire/qubes-pw-module.c
```diff
@@ -147,6 +147,31 @@ PW_LOG_TOPIC_STATIC(mod_topic, "mod." NAME);
...
+        return -EINVAL;
+    } else if (value > max_value) {
+        pw_log_error("Invalid %s \"%s\": exceeds maximum %s %zu",
+                msg, str, msg, max_value);
+        return -ERANGE;
+    } else {
+        *res = (size_t)value;
+        return 0;
+    }
+}
```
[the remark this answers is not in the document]
fepitre: This is to extract it mostly from the whole code exactly for the comment made by the original author: putting it elsewhere, and I don't think we would remind it easily. Maybe creating an issue?

pipewire/qubes-pw-module.c
```diff
@@ -835,6 +913,65 @@ static void vchan_ready(struct spa_source *source)
...
+                      const void *data,
+                      size_t size,
+                      void *user_data)
+{
+    struct impl *impl = user_data;
+    int qdb_fd;
+
+    qdb_watch(impl->qdb, QUBES_AUDIOVM_QUBESDB_ENTRY);
+    qdb_fd = qdb_watch_fd(impl->qdb);
+    spa_assert_se(qdb_fd != -EPIPE);
```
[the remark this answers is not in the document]
fepitre: I don't get it?

pipewire/qubes-pw-module.c
```diff
@@ -960,8 +960,11 @@ static void qdb_cb(struct spa_source *source)
 static int add_qdb_cb(struct spa_loop *loop, struct impl *impl)
 {
     int qdb_fd;
+    if (!qdb_watch(impl->qdb, QUBES_AUDIOVM_QUBESDB_ENTRY)) {
+        pw_log_error("Failed to setup watch on %s\n", QUBES_AUDIOVM_QUBESDB_ENTRY);
+        return -EINVAL;
```
DemiMarie, line 965:
```suggestion
        int err = -errno;
        pw_log_error("Failed to setup watch on %s: %m\n", QUBES_AUDIOVM_QUBESDB_ENTRY);
        return err;
```

DemiMarie (commented):
yes, error handling in C is terrible

marmarek (changes requested):
If audiovm isn't set initially, or not running yet, then `pipewire` fails to start at all, with `mod.qubes-audio: no /qubes-audio-domain-xid entry in QubesDB`, so it doesn't see it set later (when audiovm is set or started).

pipewire/qubes-pw-module.c
```diff
@@ -1544,43 +1665,15 @@ int pipewire__module_init(struct pw_impl_module *module, const char *args)
...
-        if (!(peer_domain_prop = pw_properties_get(props, QUBES_AUDIOVM_PW_KEY))) {
-            pw_log_error("Failed to set %s key - out of memory?", QUBES_AUDIOVM_PW_KEY);
-            res = -ENOMEM;
-            goto error;
-        }
+    qdb_handle_t qdb = qdb_open(NULL);
+    if (!qdb) {
+        res = -errno;
+        pw_log_error("Could not open QubesDB");
+        goto error;
```
DemiMarie, line 1672: If the peer is set in props, there is no need to open qubesdb.

DemiMarie (changes requested):

marmarek:
Relevant audiovm tests have passed :)

fepitre:
@DemiMarie  @marmarek  Are we good for this PR?

DemiMarie:
@fepitre good from my side

pipewire/qubes-pw-module.c
```diff
@@ -835,6 +910,67 @@ static void vchan_ready(struct spa_source *source)
...
+    domid = get_domid_from_qdb(impl->qdb);
+
+    if (domid != impl->domid) {
+        if (domid >= 0)
+            pw_log_info("Setting new peer domain ID %d", domid);
+        else if (domid == -ENOENT) // not an error, AudioVM unset
+            pw_log_info("AudioVM unset, disconnecting from %d", impl->domid);
+        else {
+            errno = -domid;
+            pw_log_error("Cannot obtain new peer domain ID: %m");
```
marmarek, line 941: Shouldn't it keep old `impl->domid` in this case?

pipewire/qubes-pw-module.c
```diff
@@ -835,6 +910,67 @@ static void vchan_ready(struct spa_source *source)
...
+        }
+        impl->domid = domid;
+        for (int i = 0; i < 2; ++i) {
+            if (impl->stream[i].source.fd != -1)
+                pw_log_error("Closing %d", impl->stream[i].source.fd);
```
marmarek, line 946: I don't think it's an error, more like info?. And also, why only log message is under this "if", is that intentional?

pipewire/qubes-pw-module.c
```diff
@@ -1691,6 +1767,27 @@ int pipewire__module_init(struct pw_impl_module *module, const char *args)
...
+        // then start the qubesdb watcher
+        if (domid_from_props<0) {
+            qdb_handle_t qdb = qdb_open(NULL);
+            if (!qdb) {
+                res = -errno;
+                pw_log_error("Could not open QubesDB");
+                goto error;
+            }
+            impl->qdb = qdb;
+            impl->domid = get_domid_from_qdb(impl->qdb);
```
marmarek, line 1782: Getting domid should be done just after setting up the watch, otherwise you'll miss modification between those two calls.

marmarek (changes requested):
Besides some minor comments, please squash those commits together.

pipewire/qubes-pw-module.c
fepitre: Because this message is not pertinent if there is no fd, we don't close -1 as this is the value saying there is no fd.

pipewire/qubes-pw-module.c
fepitre: This is going to be either a negative value or an acceptable value. It will change in all the case on new audiovm so I don't think it matters.

marmarek: Ok, so you interpret failing to read as "no audiovm". I'm fine with this, but please adjust the message to be clear about it.

marmarek (approved):

PR #214, opened 2024-06-18
Force use of software rendering

ben-grande:
Fixes: https://github.com/QubesOS/qubes-issues/issues/9268

---

Awaiting reply to https://github.com/QubesOS/qubes-issues/issues/9268#issuecomment-2175591383

### DIFF
--- a/appvm-scripts/etc/profile.d/qubes-gui.csh
+++ b/appvm-scripts/etc/profile.d/qubes-gui.csh
@@ -1,2 +1,7 @@
 setenv DISPLAY ":0"
 setenv _JAVA_AWT_WM_NONREPARENTING "1"
+if ( -f /var/run/qubes-service/software-rendering )
+  setenv GSK_RENDERER "cairo"
+  setenv GDK_DEBUG "gl-disable vulkan-disable"
+  setenv LIBGL_ALWAYS_SOFTWARE "1"
+endif
--- a/appvm-scripts/etc/profile.d/qubes-gui.sh
+++ b/appvm-scripts/etc/profile.d/qubes-gui.sh
@@ -1,1 +1,5 @@
 export DISPLAY=:0 _JAVA_AWT_WM_NONREPARENTING=1
+if test -f /var/run/qubes-service/software-rendering; then
+  export GSK_RENDERER="cairo" GDK_DEBUG="gl-disable vulkan-disable" \
+         LIBGL_ALWAYS_SOFTWARE=1
+fi


rustybird:
There's also [`GDK_DEBUG=gl-disable`](https://github.com/QubesOS/qubes-issues/issues/7921#issuecomment-1409499825) which I've been using for a while. Would it make sense to set that as well? It seems to be similar to `GSK_RENDERER=cairo` but I don't know how the two are related, maybe one covers more situations than the other.

ben-grande:
Using only `GDK_DEBUG=gl-disabled` works but it is slow because it tries vulkan:
```
$ env GDK_DEBUG="gl-disable" LIBGL_ALWAYS_SOFTWARE=1 zenity --question --title title --text text --timeout 1

Gsk-Message: Failed to realize renderer of type 'GskNglRenderer' for surface 'GdkX11Toplevel': GL support disabled via GDK_DEBUG

Gsk-Message: Failed to realize renderer of type 'GskGLRenderer' for surface 'GdkX11Toplevel': GL support disabled via GDK_DEBUG


(zenity:2633): Gdk-CRITICAL **: Vulkan: ../src/imagination/vulkan/pvr_device.c:854: Failed to enumerate drm devices (errno 2: No such file or directory) (VK_ERROR_INITIALIZATION_FAILED)

(zenity:2633): Gdk-CRITICAL **: Vulkan: Loader Message: setup_loader_term_phys_devs: Call to 'vkEnumeratePhysicalDevices' in ICD /usr/lib64/libvulkan_powervr_mesa.so failed with error code -3

(zenity:2633): Gdk-CRITICAL **: Vulkan: ../src/imagination/vulkan/pvr_device.c:854: Failed to enumerate drm devices (errno 2: No such file or directory) (VK_ERROR_INITIALIZATION_FAILED)

(zenity:2633): Gdk-CRITICAL **: Vulkan: Loader Message: setup_loader_term_phys_devs: Call to 'vkEnumeratePhysicalDevices' in ICD /usr/lib64/libvulkan_powervr_mesa.so failed with error code -3

(zenity:2633): Gdk-CRITICAL **: Vulkan: ../src/imagination/vulkan/pvr_device.c:854: Failed to enumerate drm devices (errno 2: No such file or directory) (VK_ERROR_INITIALIZATION_FAILED)

(zenity:2633): Gdk-CRITICAL **: Vulkan: Loader Message: setup_loader_term_phys_devs: Call to 'vkEnumeratePhysicalDevices' in ICD /usr/lib64/libvulkan_powervr_mesa.so failed with error code -3

(zenity:2633): Gdk-CRITICAL **: Vulkan: ../src/imagination/vulkan/pvr_device.c:854: Failed to enumerate drm devices (errno 2: No such file or directory) (VK_ERROR_INITIALIZATION_FAILED)

(zenity:2633): Gdk-CRITICAL **: Vulkan: Loader Message: setup_loader_term_phys_devs: Call to 'vkEnumeratePhysicalDevices' in ICD /usr/lib64/libvulkan_powervr_mesa.so failed with error code -3
```

With vulkan disabled also, it takes half the time of the previous example:
```sh
$ env GDK_DEBUG="gl-disable vulkan-disable" LIBGL_ALWAYS_SOFTWARE=1 zenity --question --title title --text text --timeout 1
Gsk-Message: Failed to realize renderer of type 'GskNglRenderer' for surface 'GdkX11Toplevel': GL support disabled via GDK_DEBUG

Gsk-Message: Failed to realize renderer of type 'GskGLRenderer' for surface 'GdkX11Toplevel': GL support disabled via GDK_DEBUG

Gsk-Message: Failed to realize renderer of type 'GskVulkanRenderer' for surface 'GdkX11Toplevel': Vulkan support disabled via GDK_DEBUG
```

@rustybird, from the thread you shared, I found the [following comment](https://gitlab.gnome.org/GNOME/nautilus/-/issues/2814#note_1723721):

> PS @marmarek although I strongly discourage this in general practice, for this situation, have you tried GSK_RENDERER=cairo nautilus
> ...
> GDK_RENDERER doesn't help because it's not a known envvar 
> GDK_DEBUG=gl-disable probably has the same effect as GSK_RENDERER=cairo, but give it a try anyways
> ...
> Remember, setting the renderer to cairo is not a recommendation.  I only suggested it for this very specific use case

Gnome developer says that setting the renderer to `cairo` or disabling `gl` and `vulkan` is not recommended. But they also didn't fix the nautilus spinner issue as a [patch for it still exists](https://github.com/QubesOS/qubes-nautilus/blob/main/0001-Disable-spinner-in-loading-message.patch).

@DemiMarie [said](https://github.com/QubesOS/qubes-issues/issues/7921#issuecomment-1409532131):
> The problem is that the alternate backend (Cairo) is not fully functional (it does not implement 3D transformations) and so enforcing its use will cause some programs to not render correctly.

So... maybe instead of fighting Gnome programs, like the [default templates changed to Xfce](https://forum.qubes-os.org/t/default-pre-installed-template-flavor-considering-change-from-gnome-to-xfce/20269), change to another tool, which I do not know if there is anything suitable that is a GUI with few dependencies that doesn't require hardware acceleration. I know it will not be good UX wise if we switch to `cdialog`, but `yad` and `kdialog` requires a lot of undesired dependencies, GTK dialog can take more time to implement as sometimes `zenity` is called from python and sometimes from shell scripts.

rustybird:
Agreed that it would be good to find an alternative to zenity. Also because the Debian build of it is so bloated (from WebKit support) that it's not even included in the minimal Debian template.

But the bad performance of modern GTK programs still seems like an important thing to address too. The only way I can see is switching to Cairo, even if it potentially breaks some programs. (Which ones though?)

Maybe it could be done through an easier to override mechanism? E.g. as a `.conf` file in `/usr/lib/environment.d` spamming all three environment variables (and including `vulkan-disable`). That way, people for whom Cairo actually breaks stuff can switch away from it.

rustybird:
> Maybe it could be done through an easier to override mechanism? E.g. as a `.conf` file in `/usr/lib/environment.d`

Ah wait, that might also interact with sys-gui-gpu :thinking:

DemiMarie:
> Agreed that it would be good to find an alternative to zenity. Also because the Debian build of it is so bloated (from WebKit support) that it's not even included in the minimal Debian template.

Indeed Zenity should not be used.  

> But the bad performance of modern GTK programs still seems like an important thing to address too. The only way I can see is switching to Cairo, even if it's not future-proof and potentially breaks some programs. (Which ones though?)

The proper and long-term solution is to expose the GPU to guests in a secure way.  Everything else is a workaround until secure GPU acceleration can be implemented.  Unfortunately, GPU security has lagged behind CPU security, but this must be fixed on the GPU side.  Software rendering is a fallback for a reason.

In the short term, switching to Cairo is one option.  I would be much more confortable doing so on a per-program basis than system-wide, though.  Otherwise there will be programs that break for no apparent reason.

> Maybe it could be done through an easier to override mechanism? E.g. as a `.conf` file in `/usr/lib/environment.d` spamming all three environment variables (and including `vulkan-disable`). That way, people for whom Cairo actually breaks stuff can switch away from it by masking the file with a symlink to `/dev/null`. Do you think that's a sensible tradeoff @DemiMarie @marmarek?

@marmarta do you have an opinion here?

ben-grande:
Just to get an idea of all components that uses zenity and how many times it appears per file:
```sh
$ rg -c -e "zenity -" -e "'zenity', '-"

qubes-app-linux-split-gpg/qubes.Gpg.service:1
qubes-app-linux-split-gpg2/splitgpg2/__init__.py:1
qubes-core-agent-linux/qubes-rpc/qubes.SelectFile:1
qubes-core-agent-linux/qubes-rpc/qvm-copy-to-vm.gnome:1
qubes-core-agent-linux/qubes-rpc/thunar/qvm-actions.sh:3
qubes-core-agent-linux/qubes-rpc/vm-file-editor.c:1
qubes-core-agent-linux/qubes-rpc/qvm-open-in-vm:1
qubes-core-agent-linux/qubes-rpc/qubes.SelectDirectory:1
qubes-core-agent-linux/debian/changelog:1
qubes-core-agent-linux/package-managers/qubes-download-dom0-updates.sh:2
qubes-core-admin-linux/file-copy-vm/qfile-dom0-agent.c:1
qubes-core-admin-linux/dom0-updates/qubes-dom0-update:2
qubes-core-admin/qubes/tests/integ/basic.py:2
qubes-app-linux-img-converter/qvm-convert-img.gnome:1
qubes-app-linux-img-converter/qimg-convert-client:1
qubes-app-linux-img-converter/qvm-convert-img-gnome:1
qubes-app-linux-pdf-converter/qvm-convert-pdf.gnome:1
qubes-core-qrexec/daemon/qrexec-daemon.c:1
```

marmarek:
> But the bad performance of modern GTK programs still seems like an important thing to address too. The only way I can see is switching to Cairo, even if it's not future-proof and potentially breaks some programs. (Which ones though?)

I'm fine with this solution, if there is an easy opt-out. Nobody shown any broken apps by this approach yet, so until that happens, I would ignore this downside as it looks purely theoretical.

As for the opt-out - maybe use qvm-service? that would allow easily doing it per-qube. And on the global scale you can always place some later file in /etc/profile.d (or whichever other dir is chosen), but it needs to be documented somewhere.

appvm-scripts/etc/profile.d/qubes-gui.sh
```diff
@@ -1 +1,5 @@
 export DISPLAY=:0 _JAVA_AWT_WM_NONREPARENTING=1
+if ! test -f /var/run/qubes-service/gpu-accel; then
```
marmarek, line 2: I'd prefer a name like `software-rendering` or similar, instead of `gpu-accel`. The latter may suggest enabling it will actually get you GPU acceleration, which is not really the case.

marmarek (changes requested):

appvm-scripts/etc/profile.d/qubes-gui.sh
ben-grande: Sure. I am searching where default services are set to enable `software-rendering` by default... Until I find them, I will commit with the same logic: ` ! test -f /var/run/qubes-service/hardware-rendering`.

ben-grande: Found, `vm-systemd/qubes-sysinit.sh`

appvm-scripts/etc/profile.d/qubes-gui.csh
```diff
@@ -1,2 +1,5 @@
-setenv DISPLAY ":0"
-setenv _JAVA_AWT_WM_NONREPARENTING "1"
+setenv DISPLAY ":0" _JAVA_AWT_WM_NONREPARENTING "1"
```
marmarek, line 1: I don't think you can set multiple variables with a single `setenv` call...

appvm-scripts/etc/profile.d/qubes-gui.csh
```diff
@@ -1,2 +1,5 @@
...
+if ( -f /var/run/qubes-service/software-rendering )
+  setenv GSK_RENDERER "cairo" GDK_DEBUG "gl-disable vulkan-disable" \
```
marmarek, line 3: And here too

marmarek (changes requested):

appvm-scripts/etc/profile.d/qubes-gui.csh
marmarek:
```
$ setenv test "abc" test2 "def"
setenv: Too many arguments.
```

marmarek (approved):

PR #216, opened 2024-09-26
Prevent $DISPLAY from being overridden in SSH (for X11 forwarding)

InnovativeInventor:
When trying to ssh into a qubes vm, `/etc/profile.d/qubes-gui.sh` will improperly set `$DISPLAY` to be :0. When X11 forwarding is used, `$DISPLAY` is automatically set to ensure that the relevant windows get forwarded to the ssh client. However, since qubes-gui.sh is called after a successful login, the correct value for `$DISPLAY` will be overridden.

This commit fixes the issue by adding a test in qubes-gui.sh to check if we are currently in an ssh session (by checking if `$SSH_CLIENT` or `$SSH_TTY` are set).

Note: I do not know .csh scripts and have not tested the proposed change to qubes-gui.csh, yet.


appvm-scripts/etc/profile.d/qubes-gui.sh
```diff
@@ -1,4 +1,6 @@
-export DISPLAY=:0 _JAVA_AWT_WM_NONREPARENTING=1
+if ! ([ -n "$SSH_CLIENT" ] || [ -n "$SSH_TTY" ] ]); then
+  export DISPLAY=:0 _JAVA_AWT_WM_NONREPARENTING=1
+fi
```
marmarek, line 3: First, this should apply only to `DISPLAY`, not the other variable.
But second, having it specifically for SSH isn't very elegant. Maybe better simply check if it's set already and not override it then? This can be done simpler:
```
: ${DISPLAY:=:0}
export DISPLAY
```

marmarek (changes requested):

PR #220, opened 2024-12-06
pipewire: prevent starting in root session

marmarek:
If root session is started early and pipewire gets started as part of
it, it may connect to the audiovm earlier than the user's instance,
effectively stealing the connection. Upstream already has
ConditionUser=!root in relevant pulseaudio units, but not in pipewire
ones. Add similar condition to pipewire.service unit.

Fixes QubesOS/qubes-issues#9619


DemiMarie:
Should there be a similar check for WirePlumber?

marmarek:
wireplumber has BindsTo=pipewire, so it should be implicit already; but even if not, just wireplumber is harmless.

PR #222, opened 2024-12-21
disable qubes-icon-sender on sys-net and sys-usb

Guiiix:
Qubes icon sender allows to update applications windows icons. In practice, few windows are opened on the sys-vms. Disabling this service allows to recover some RAM without causing inconvenience.

### DIFF
--- a/window-icon-updater/qubes-icon-sender.desktop
+++ b/window-icon-updater/qubes-icon-sender.desktop
@@ -8,3 +8,4 @@ Type=Application
 Categories=
 GenericName=
 OnlyShowIn=X-QUBES;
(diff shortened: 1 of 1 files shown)


marmarek:
PipelineRetry

marmarek (approved):

ben-grande:
Took me a minute but for others that are reading and trying to understand why `Exec=/usr/lib/qubes/qvm-service-wrapper program` was not used instead, it is because `qubes-icon-sender` is not in `$PATH` so can't use basename which is necessary for the detection of `/run/qubes-service/$service` to work. Naming the service `qubes-icon-sender` and enabling by default in Qubes could be another solution, but as I see the minimal-(netvm|usbvm) being used elsewhere, I understand the reasoning now.

PR #224, opened 2025-01-22
Pull in required Pipewire dependencies for Archlinux

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9660
Reference: https://github.com/QubesOS/qubes-meta-packages/pull/75

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,5 +12,5 @@ include:
   project: QubesOS/qubes-continuous-integration
   ref: $CI_BRANCH_REF
 variables:
-  CI_INSTALL_EXCLUDE: pulseaudio-qubes pulseaudio-qubes-dbgsym
+  CI_INSTALL_EXCLUDE: pulseaudio-qubes pulseaudio-qubes-dbgsym qubes-vm-pulseaudio
   CI_INSTALL_EXCLUDE_DOM0: pulseaudio-qubes qubes-gui-agent
--- a/archlinux/PKGBUILD.in
+++ b/archlinux/PKGBUILD.in
@@ -16,7 +16,7 @@ makedepends=(
     automake
     autoconf
     libtool
-    'pulseaudio<=17.0+r43+g3e2bb8a1e'
+    'libpulse<=17.0+r43+g3e2bb8a1e'
     'pipewire>=1:0.3.53-2'
     xorg-server-devel
     xorg-util-macros
@@ -102,7 +102,7 @@ package_qubes-vm-pulseaudio() {
 package_pipewire-qubes () {
 
     pkgdesc='PipeWire support for Qubes VM.  This replaces the legacy PulseAudio module.'
-    depends=('pipewire>=1:0.3.53-2')
+    depends=('pipewire>=1:0.3.53-2' 'pipewire-pulse' 'wireplumber')
     replaces=('qubes-vm-pulseaudio')
     install=archlinux/PKGBUILD-pipewire-qubes.install
 


marmarek:
no comas

marmarek:
and also, is pipewire-alsa really needed?

alimirjamali:
> and also, is pipewire-alsa really needed?

I excluded `pipewire-jack` as it is seldom used. But pipewire-alsa might be used by more users.

marmarek:
The thing is, with pipewire-qubes you talk to dom0 (or other audiovm), not to devices directly. pipewire-alsa might be relevant for sys-audio, but there you don't really need pipewire-qubes, so it isn't the best place to put such dependency

alimirjamali:
> The thing is, with pipewire-qubes you talk to dom0 (or other audiovm), not to devices directly. pipewire-alsa might be relevant for sys-audio, but there you don't really need pipewire-qubes, so it isn't the best place to put such dependency

Well said. I removed `pipewire-alsa` from the PR.

marmarek:
CI says:
> 14:02:10 [qb.gitlab-installv2] resolving dependencies...
14:02:10 [qb.gitlab-installv2] error: unresolvable package conflicts detected
14:02:10 [qb.gitlab-installv2] error: failed to prepare transaction (conflicting dependencies)
14:02:10 [qb.gitlab-installv2] looking for conflicting packages...
14:02:10 [qb.gitlab-installv2] :: pipewire-pulse-1:1.2.7-1 and pulseaudio-17.0+r43+g3e2bb8a1e-1 are in conflict

It looks like something more is needed for clean upgrade. Maybe add conflict with pulseaudio here too? Or add pulseaudio to `replaces` list? I'm not exactly sure about semantics of various dependencies types in Arch - besides clean upgrade, this also should avoid unintended side effects like pretending that pipewire-qubes provides pulseaudio...

alimirjamali:
> Or add pulseaudio to `replaces` list?

I did this, hopefully it solves the issue this time.

marmarek:
CI here still complains...

alimirjamali:
> CI here still complains...

Shouldn't we expect it? Since the [other change](https://github.com/QubesOS/qubes-meta-packages/pull/75/files) for meta-packages is still not merged. And `qubes-vm-dependencies` still depends on `qubes-vm-pulseaudio`. Which itself depends on `pulseaudio`.

marmarek:
I don't think qubes-vm-dependencies is installed in this test docker...

marmarek:
It's more or less plain `archlinux:latest` docker: https://gitlab.com/QubesOS/qubes-continuous-integration/-/blob/main/dockerfiles/vm-archlinux.Dockerfile?ref_type=heads

alimirjamali:
1st of all, I am not an Archlinux packaging expert, or a Docker expert. I am sorry if I am making more or less wild guesses here.

2nd of all, in [makedepends](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/archlinux/PKGBUILD.in#L19-L20), I see both pulseaudio and pipewire are installed? Is this normal? Why requiring the main packages for building rather than their libraries.

marmarek:
> 2nd of all, in [makedepends](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/archlinux/PKGBUILD.in#L19-L20), I see both pulseaudio and pipewire are installed? Is this normal? Why requiring the main packages for building rather than their libraries.

This package include agents for both pipewire and pulseaudio - in separate sub-packages. But yes, only libraries should be needed, if they are in separate packages in Arch.

That said, maybe the CI issue is more fundamental here - I think it tries to install all subpackages at once, which will include both pipewire and pulseaudio agents... I see there is CI_INSTALL_EXCLUDE set to pulseaudio-qubes here, but the Arch package is named qubes-vm-pulseaudio...

marmarek:
Try adding qubes-vm-pulseaudio to CI_INSTALL_EXCLUDE in .gitlab-ci.yml

alimirjamali:
> Try adding qubes-vm-pulseaudio to CI_INSTALL_EXCLUDE in .gitlab-ci.yml

Ok. CI Tests are also green. I also replaced `pulseaudio<=17.0+r43+g3e2bb8a1e` with `
libpulse<=17.0+r43+g3e2bb8a1e` in the make dependencies. And assured `qubes-vm-pulseaudio` would also replace `pipewire-qubes`.

The OpenQA tests will be performed in the background?

marmarek:
openQA complains: https://openqa.qubes-os.org/tests/126666#step/update2/64
So, automatic update doesn't work. I guess interactive would, if one would answer "y" to that prompt...

marmarek:
I confirm that manual update works, if one answers "y" for the question of removing pulseaudio. But then, on the next update, pacman proposes to switch back to qubes-vm-pulseaudio!

alimirjamali:
PipelineRetryFailed

alimirjamali:
> I confirm that manual update works, if one answers "y" for the question of removing pulseaudio. But then, on the next update, pacman proposes to switch back to qubes-vm-pulseaudio! https://openqa.qubes-os.org/tests/126716#step/update_post_templates/26

@marmarek The last OpenQA report is from yesterday. Is it after or before my last amendment?

marmarek:
> @marmarek The last OpenQA report is from yesterday. Is it after or before my last amendment?

tl;dr before

You can always check by looking at the top of the report - list of the PRs include link to exact commit that was tested. Similarly, when looking that openQA job, in the "settings" tab the PULL_REQUEST entry has exact commit ID that is being tested for each PR.

marmarek:
ok, I think I found a solution, ugly one, but works:
Before the update run `pacman -Rdd pulseaudio`. This removes pulseaudio ignoring dependencies. And the update will install pipewire-pulse. I haven't found any method to do that in a single operation, which is not ideal, since if the update fails, it will leave the template without either pulseaudio backend... But then, re-running the update should fix it, so I guess it's okay-ish.

marmarek (approved):

PR #225, opened 2025-02-25
xf86-video-dummy: fix type of CreateWindow field

marmarek:
The function takes an argument. Actual function used there was okay, but
the field in the structure used wrong type. And GCC 15 (rightfully)
started complaining.
Use CreateWindowProcPtr type to avoid similar issue in the future.

QubesOS/qubes-issues#9807


marmarek:
PipelineRetry

PR #231, opened 2025-05-27
Send icons for legacy X11 Apps

alimirjamali:
Since very old X11 Apps (Xterm, Xcalc, Xlogo, Xclock, Xeyes, ...) do not have the `_NET_WM_ICON` Atom, revert to `WM_HINT` and extract icons and their transparency with that technology. This is mostly useful for Xterm

fixes: https://github.com/QubesOS/qubes-issues/issues/9973

### DIFF
--- a/window-icon-updater/icon-sender
+++ b/window-icon-updater/icon-sender
@@ -33,11 +33,10 @@ import time
 
 import xcffib
 from xcffib import xproto
-from Xlib.display import Display
-from Xlib import X
-from Xlib.error import XError
 
 ICON_MAX_SIZE = 256
+IconPixmapHint = 0b1 << 2
+IconMaskHint = 0b1 << 5
 
 
 log = logging.getLogger('icon-sender')
@@ -52,7 +51,6 @@ class IconRetriever(object):
         self.conn = xcffib.connect()
         self.setup = self.conn.get_setup()
         self.root = self.setup.roots[0].root
-        self.display = Display()
 
         # just created windows for which icon wasn't sent yet - should
         # be send on MapNotifyEvent
@@ -64,7 +62,6 @@ class IconRetriever(object):
     def disconnect(self):
         log.info('disconnecting from X')
         self.conn.disconnect()
-        self.display.close()
 
     def watch_window(self, w):
         self.conn.core.ChangeWindowAttributesChecked(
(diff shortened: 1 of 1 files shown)


window-icon-updater/icon-sender
```diff
@@ -49,6 +52,7 @@ class IconRetriever(object):
         self.conn = xcffib.connect()
         self.setup = self.conn.get_setup()
         self.root = self.setup.roots[0].root
+        self.display = Display()
```
marmarek, line 55: Don't make the second X11 connection with another library, use XCB connection that is here already.

marmarek (changes requested):

window-icon-updater/icon-sender
```diff
@@ -82,7 +84,134 @@ class IconRetriever(object):
...
+                        0,
+                        icon_mask_geometry.width,
+                        icon_mask_geometry.height,
+                        0xFFFFFFFF
+                    ).reply()
+                    icon_mask_data = image.data.raw
+            except (
+                    xcffib.xproto.WindowError,
+                    xcffib.xproto.AccessError,
+                    NoIconError
```
marmarek, line 147: You don't need to catch `NoIconError` if you're going to transform it into the same exception.

window-icon-updater/icon-sender
```diff
@@ -82,7 +84,134 @@ class IconRetriever(object):
...
+                        byte = int(icon_mask_data[byte_offset])
+                        byte = byte >> (x % 8)
+                        bit = byte & 0x1
+                        pixel = x + y * icon_mask_geometry.height
+                        if bit:
+                            icon_data[pixel] = icon_data[pixel] & 0xffffffff
+                        else:
+                            icon_data[pixel] = icon_data[pixel] & 0x00ffffff
+            elif icon_mask_data and icon_mask_geometry.depth == 8:
+                # Technically this is not tested (No X prog uses 8bit/pix mask)
```
marmarek, line 201: Indeed... `row_width` is not set but is used in this case...

window-icon-updater/icon-sender
```diff
@@ -82,7 +84,134 @@ class IconRetriever(object):
...
+                # At least not on Qubes OS 4.3 & default Xfwm4
+                for y in range(0, icon_mask_geometry.height):
+                    offset = y * row_width
+                    for x in range(0, icon_mask_geometry.width):
+                        byte_offset = x + offset
+                        byte = int(icon_mask_data[byte_offset])
+                        pixmask = (byte < 24) | 0x00ffffff
```
marmarek, line 208:
```suggestion
                        pixmask = (byte << 24) | 0x00ffffff
```

window-icon-updater/icon-sender
alimirjamali: This is done

window-icon-updater/icon-sender
alimirjamali: This is done

window-icon-updater/icon-sender
alimirjamali: A changed this a little bit, catching another error (`xproto.WindowError`).

window-icon-updater/icon-sender
```diff
@@ -82,7 +84,133 @@ class IconRetriever(object):
...
+                        0,
+                        icon_mask_geometry.width,
+                        icon_mask_geometry.height,
+                        0xFFFFFFFF
+                    ).reply()
+                    icon_mask_data = image.data.raw
+            except (xproto.BadWindow, xproto.WindowError, xproto.AccessError):
+                raise NoIconError
+            except Exception:   # pylint: disable=bare-except
+                raise NoIconError
```
marmarek, line 147: What exactly falls into this? Silently ignoring all other errors is a bad idea. At the very least add a log message, but you can also simply drop this one, to not continue in case of fatal error (like connection to X server lost).

marmarek (changes requested):

window-icon-updater/icon-sender
marmarek: > But I tried the code without it and it fails

Then not ignoring that exception is even more important to find out what is wrong. Maybe `traceback.print_exc()` will help? And/or printing `repr(e)` or os.

> But it gives no clue (It only catch `NoIconError` errors).

`NoIconError` should _not_ be handled here, it should be handled by the caller.

alimirjamali: Just remembered how to reproduce the bug. It happens if you install `xlogo` in Fedora without `xorg-x11-xbitmaps` & `xorg-x11-fonts-75dpi` packages. And they you try to run `xlogo`. In that case, it crashes it (the crash happens within the try/except.

xlogo prints "Warning: Cannot convert string "xlogo32" to type Pixmap"

While this is a failure of Fedora packaging of setting the dependencies, still the code should be resilient.

alimirjamali: Found it. Additional error to catch is `xcffib.xproto.DrawableError`. I will amend the PR in a minute.

marmarek (approved):

PR #233, opened 2025-06-08
Do not use grant tables for allocating framebuffer

marmarek:
It seems like the framebuffer is never shared with the GUI daemon directly
(at least without GPU passthrough), so it doesn't need to be backed by grant tables. Allocating it as normal userspace memory has several benefits:
- it can be swapped if needed
- kernel has more options to allocate it (it's no longer allocated from kernel
  memory)
- it can be resized later, allowing dynamic videoram changes (no need to restart VMs after connecting external display anymore!)

There is one corner case I'm not sure about when an actual GPU is available and
glamor is used. So, in this case I keep the old behavior.
See commit messages for more details.

@HW42 can you sanity check this? You added grant tables support initially, so
maybe you remember why it was done this way...

Fixes QubesOS/qubes-issues#7448
Fixes QubesOS/qubes-issues#9992


marmarek:
> Do we have someone that uses the glamor setup (I see it was added around 2021)? Then we could let them test drive it, to be really sure.

It was added together (and for) GVT-g support, and the most important part (dom0 kernel patch) is not in our repos, as it was too much risk. That technology is basically dead. Theoretically glamor should work also with any GPU passthrough, and did managed to get it initialized (at least according to X server logs), but I'm not really sure how to test it properly. I can tell it didn't crash (even without limiting this change to non-glamor case), but I can't be sure it actually exercised all possible paths.

marmarek:
> Hmm, on openQA the guivm tests failed due to qubes-menu dying with `Request refused` (likely unrelated). So this should be fixed and re-run.

Most recent yes, but earlier did pass (see edits history of that report, especially the initial comment)

marmarek:
> I would vote for dropping it.

Ok, lets do it then.

marmarek:
Then the last question is about `xf86_qubes_pixmap_set_private()` called for the root window (in `qubes_create_screen_resources()`). As you noted, there is currently no way to share part of that buffer, so it's unlikely need grant tables. So, I disabled grant-based FBBase even for the glamor case, but left the code in place just in case (it's just a few lines so no big deal).

marmarek:
Ugh, now I found this in one of VMs:
```
[     9.511] (EE) can't dump window without grant table allocation
[     9.517] (EE) can't dump window without grant table allocation
[ 55848.651] (EE) can't dump window without grant table allocation
```
I haven't noticed any breakage, like missing content of some window, but nevertheless it tried to dump content of a window without grants. OTOH, it appears as at least one of those messages are related to this window:
```
     0x160001b "[redacted]: 451x576+734+1752  +734+1752
```
which definitely is not the root window. Maybe unrelated issue? Some race condition or something?

PR #236, opened 2025-06-21
Don't try to send grants for not realized windows

HW42:
We rely on composite redirect mode of the X server to get per window pixmaps. Those are setup/teared-down in compRealizeWindow/ compUnrealizeWindow (via compCheckRedirect).

So not realized windows don't have a per window pixmap. Sending grant refs for them was always broken since we didn't send the offset into the screen pixmap in those cases. But with the recent change to not allocate grant refs for the screen pixmap this leads to a noticable error message.

So don't try to send grant refs for not realized windows. This means that the configure before mapping will not contiain grant refs. But when we map the window we will get a damage event because of the new pixmap compRealizeWindow has allocated and send them then. So this should be fine.

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -159,14 +159,38 @@ static struct genlist *windows_list;
 static struct genlist *embeder_list;
 static Ghandles *ghandles_for_vchan_reinitialize;
 
-#define SKIP_NONMANAGED_WINDOW do {                                    \
-    if (!list_lookup(windows_list, window)) {                          \
-        if (g->log_level > 0)                                          \
-            fprintf(stderr, "Skipping unmanaged window 0x%lx\n",       \
-                    window);                                           \
-        return;                                                        \
-    }                                                                  \
-} while (0)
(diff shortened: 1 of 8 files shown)


gui-agent/vmside.c
```diff
@@ -898,10 +901,14 @@ static void process_xevent_map(Ghandles * g, XID window)
 static void process_xevent_unmap(Ghandles * g, XID window)
 {
     struct msg_hdr hdr;
+    struct window_data *wd;
     SKIP_NONMANAGED_WINDOW;
 
+    wd = list_lookup(windows_list, window)->data;
```
marmarek, line 907: Since you do list_lookup here anyway, you can open-code (modified) `SKIP_NONMANAGED_WINDOW` here to avoid traversing the list twice.

marmarek (changes requested):

marmarek:
Interesting, so it seems we found this issue before already, just hasn't realized its full scope: 8b53d361810ced38974378b386bd215b0a52d6e8

marmarek:
(none of the openQA failures is related to this PR)

marmarek (approved):

PR #251, opened 2025-11-14
Delay connecting with gui-daemon until after starting Xorg

marmarek:
Until now, Xorg was started only after connecting with gui-daemon and
receiving xconf message. This was mostly because initial
resolution/video mem specified in the xorg.conf couldn't later be
changed and that affects maximum working resolution.

Since 49420c5 "Increase screen pixmap size beyond initial video RAM if
needed", this is no longer a problem. The only parameter (still) needed
before starting Xorg is GUI domain id, but this is available on the
command line. Change gui-agent to start Xorg earlier and only then wait
for gui-daemon. This allows starting all of the user session earlier,
optimizing startup time. This is especially relevant for VMs started on
boot (before user logs in), when GUI daemon isn't started yet. And also
helpful for preloaded disposables, which may start user session as part
of preloading now.

The current implementation is rather naive: it starts Xorg with
hardcoded 1920x1080 resolution (which will later be updated by the
qubes.SetMonitorLayout qrexec call), selects which events it want to
receive and then waits for the GUI daemon. Especially, no events are
actually processed before GUI daemon connects. This assumes all events
will be queued and can be processed after GUI daemon connection. Very
similar approach is already taken on GUI daemon re-connection, and it
works.

There is a potential issue, if too many events get queued before GUI
daemon connects and some get dropped. It's unclear how Xlib handles that
(and what is the limit), but if that happens there are two alternative
solutions:

1. Start processing events normally, especially collect info about all
relevant windows, but dont send anything to GUI daemon before it
connects - this basically requires changing write_struct() (and other
similar places) to check if GUI daemon is connected. And once GUI daemon
connects (first event received on the vchan FD), call
send_all_windows_info().

2. Wait for GUI daemon before registering for events, but then iterate
over all windows to collect necessary info and send it to dom0. This
approach theoretically can be more reliable (and might even allow
restarting gui-agent without restarting Xorg as a side effect), but it's
unclear if all data can be rebuilt this way. Especially tray icons may
be problematic, as it isn't any window property, but a message sent to
relevant selection owner (but theoretically
https://tronche.com/gui/x/icccm/sec-2.html#s-2.8 provides
solution for this issue).

https://github.com/QubesOS/qubes-issues/issues/9940#issuecomment-2847640715

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,6 +168,8 @@ install-common:
 	install -d $(DESTDIR)/etc/qubes/post-install.d
 	install -m 0755 appvm-scripts/etc/qubes/post-install.d/20-qubes-guivm-gui-agent.sh \
                 $(DESTDIR)/etc/qubes/post-install.d/20-qubes-guivm-gui-agent.sh
+	install -m 0755 appvm-scripts/etc/qubes/post-install.d/20-qubes-gui-agent.sh \
+                $(DESTDIR)/etc/qubes/post-install.d/20-qubes-gui-agent.sh
 	install -D appvm-scripts/usrbin/qubes-session \
 		$(DESTDIR)/usr/bin/qubes-session
 	install -D appvm-scripts/usrbin/qubes-run-xorg \
--- a/appvm-scripts/etc/qubes/post-install.d/20-qubes-gui-agent.sh
+++ b/appvm-scripts/etc/qubes/post-install.d/20-qubes-gui-agent.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# announce support for late gui-daemon connection
+qvm-features-request supported-feature.late-gui-daemon=1
--- a/appvm-scripts/usrbin/qubes-run-xorg
+++ b/appvm-scripts/usrbin/qubes-run-xorg
@@ -26,7 +26,14 @@
 
 ## This script is triggered by qubes-gui-agent systemd service.
 
-#expects W, H, MEM and DEPTH env vars to be set by invoker
+#expects W, H, MEM and DEPTH env vars to be set by invoker, but use default
+# values if they aren't
+
+[ -n "$W" ] || W=1920
+[ -n "$H" ] || H=1080
+[ -n "$MEM" ] || MEM=$(( W * H * 4 / 1024 ))
+[ -n "$DEPTH" ] || DEPTH=24
+
 DUMMY_MAX_CLOCK=300 #hardcoded in dummy_drv
 PREFERRED_HSYNC=50
 RES="$W"x"$H"
--- a/debian/qubes-gui-agent.install
+++ b/debian/qubes-gui-agent.install
@@ -10,6 +10,7 @@ etc/security/limits.d/90-qubes-gui.conf
 etc/qubes-rpc/qubes.SetMonitorLayout
 etc/qubes-rpc/qubes.GuiVMSession
 etc/qubes/post-install.d/20-qubes-guivm-gui-agent.sh
+etc/qubes/post-install.d/20-qubes-gui-agent.sh
 etc/xdg/autostart/qubes-gtk4-workarounds.desktop
 etc/xdg/autostart/qubes-icon-sender.desktop
 etc/xdg/autostart/qubes-qrexec-fork-server.desktop
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -2176,12 +2176,11 @@ static void handle_close(Ghandles * g, XID winid)
 
 /* start X server, returns its PID
  */
-static pid_t do_execute_xorg(
-        char *w_str, char *h_str, char *mem_str, char *depth_str,
-        char *gui_domid_str)
+static pid_t do_execute_xorg(Ghandles *g)
 {
     pid_t pid;
     int fd;
+    char gui_domid_str[12];
 
     pid = fork();
     switch (pid) {
@@ -2189,10 +2188,7 @@ static pid_t do_execute_xorg(
             perror("fork");
             return -1;
         case 0:
-            setenv("W", w_str, 1);
-            setenv("H", h_str, 1);
-            setenv("MEM", mem_str, 1);
-            setenv("DEPTH", depth_str, 1);
+            snprintf(gui_domid_str, sizeof(gui_domid_str), "%u", g->domid);
             setenv("GUI_DOMID", gui_domid_str, 1);
             /* don't leak other FDs */
             for (fd = 3; fd < 256; fd++)
@@ -2374,27 +2370,11 @@ static void handle_message(Ghandles * g)
     }
 }
 
-static pid_t get_xconf_and_run_x(Ghandles *g)
-{
-    struct msg_xconf xconf;
-    char w_str[12], h_str[12], mem_str[12], depth_str[12], gui_domid_str[12];
-    pid_t x_pid;
-    read_struct(g->vchan, xconf);
-    snprintf(w_str, sizeof(w_str), "%d", xconf.w);
-    snprintf(h_str, sizeof(h_str), "%d", xconf.h);
-    snprintf(mem_str, sizeof(mem_str), "%d", xconf.mem);
-    snprintf(depth_str, sizeof(depth_str), "%d", xconf.depth);
-    snprintf(gui_domid_str, sizeof(gui_domid_str), "%u", g->domid);
-    x_pid = do_execute_xorg(w_str, h_str, mem_str, depth_str, gui_domid_str);
-    if (x_pid == (pid_t)-1) {
-        errx(1, "X server startup failed");
-    }
-    return x_pid;
-}
-
 static void handshake(Ghandles *g)
 {
     uint32_t version = PROTOCOL_VERSION;
+    struct msg_xconf xconf;
+
     write_struct(g->vchan, version);
     version = 0;
     read_struct(g->vchan, version);
@@ -2407,12 +2387,14 @@ static void handshake(Ghandles *g)
                 PROTOCOL_VERSION_MAJOR, PROTOCOL_VERSION_MINOR,
                 major_version, minor_version);
     g->protocol_version = version;
+
+    /* discard */
+    read_struct(g->vchan, xconf);
 }
 
 static void handle_guid_disconnect(void)
 {
     Ghandles *g = ghandles_for_vchan_reinitialize;
-    struct msg_xconf xconf;
 
     if (!ghandles_for_vchan_reinitialize) {
         fprintf(stderr, "gui-daemon disconnected before fully initialized, "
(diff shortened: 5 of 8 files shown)


marmarek:
@ben-grande I've looked at gui-agent and I think this should be enough to start user session without waiting for GUI daemon. See above description for specific details.

Generally, this should allow using `qubes.WaitForSession` even if gui-daemon isn't started yet.

This is only lightly tested, needs checking for various corner cases. A few cases that are worth testing:
1. Start VM, but not gui-daemon, start some GUI apps, and only then start gui-daemon - app windows should appear.
2. As with 1, but start and stop an app several times before starting gui-daemon (while an app window may flicker at gui-daemon startup, nothing should crash).
3. Start something with tray icon (both before starting gui-daemon and after)
4. Restart gui-daemon (logout and login to dom0?)
5. Move/resize app window before starting gui-daemon (via `qvm-run ... xdotool ...`?)
6. And ofc, try preloading with using `qubes.WaitForSession` instead of `qubes.WaitForRunningSystem` (but still start gui-daemon only when preloaded is used)

When  I write "start gui-daemon", I mean the whole thing that qvm-start-daemon does, including qubes.SetMonitorLayout call.

There are probably some more cases to test, I hope you'll think of something :)

ben-grande:
> @ben-grande I've looked at gui-agent and I think this should be enough to start user session without waiting for GUI daemon. See above description for specific details.
> 
> Generally, this should allow using `qubes.WaitForSession` even if gui-daemon isn't started yet.
> 
> This is only lightly tested, needs checking for various corner cases. A few cases that are worth testing:
> 
>     1. Start VM, but not gui-daemon, start some GUI apps, and only then start gui-daemon - app windows should appear.
> 
>     2. As with 1, but start and stop an app several times before starting gui-daemon (while an app window may flicker at gui-daemon startup, nothing should crash).
> 
>     3. Start something with tray icon (both before starting gui-daemon and after)
> 
>     4. Restart gui-daemon (logout and login to dom0?)
> 
>     5. Move/resize app window before starting gui-daemon (via `qvm-run ... xdotool ...`?)
> 
>     6. And ofc, try preloading with using `qubes.WaitForSession` instead of `qubes.WaitForRunningSystem` (but still start gui-daemon only when preloaded is used)
> 
> 
> When I write "start gui-daemon", I mean the whole thing that qvm-start-daemon does, including qubes.SetMonitorLayout call.
> 
> There are probably some more cases to test, I hope you'll think of something :)

Tested 1, 4 and lightly 6.

To test, I've downloaded the package from this PR and installed it in a template. I killed the qvm-start-daemon on login and manually started it to get some feedback when it connects to the qube as well as allowing restarting it without having to do login and logout. I haven't found issues so far, which is great, deeper test is needed and I will  try to finish testing tomorrow to hopefully get this merged before rc4.

Sometimes I wish all qubes `/etc/xdg/autostart` where systemd services so it logs and can be easily restarted, but that is another story.

ben-grande:
It is taking some time to preload, especially `qubes.WaitForSession`:
```
Nov 29 00:28:03 dom0 qubesd[53744]: INFO: vm.disp6524: Activating qube
Nov 29 00:28:03 dom0 runuser[54664]: pam_unix(runuser:session): session opened for user user(uid=1000) by (uid=0)
Nov 29 00:28:07 dom0 runuser[54664]: pam_unix(runuser:session): session closed for user user
Nov 29 00:28:07 dom0 qubesd[53744]: INFO: vm.disp6524: Preload startup waiting 'qubes.WaitForRunningSystem' with '60' seconds timeout
Nov 29 00:28:19 dom0 qubesd[53744]: INFO: vm.disp6524: Preload startup completed 'qubes.WaitForRunningSystem'
Nov 29 00:28:19 dom0 qubesd[53744]: INFO: vm.disp6524: Preload startup waiting 'qubes.WaitForSession' with '60' seconds timeout
Nov 29 00:28:53 dom0 qubesd[53744]: INFO: vm.disp6524: Preload startup completed 'qubes.WaitForSession'
Nov 29 00:28:53 dom0 qubesd[53744]: INFO: vm.disp6524: Setting qube memory to pref mem
Nov 29 00:28:56 dom0 qubesd[53744]: WARNING: vm.disp6524: Failed to set memory
Nov 29 00:28:56 dom0 qubesd[53744]: INFO: vm.disp6524: Preloading finished
Nov 29 00:28:56 dom0 qubesd[53744]: INFO: vm.disp6524: Paused preloaded qube
```

Unsure yet what is taking that long for the user session to start, it is taking 30 seconds... while system is ready very fast. 

qubes-notification-agent shows `2.7` seconds on `systemd-analyze --user blame`, but there is nothing else taking a lot of seconds.

sys
user

---

Measuring raw call.

```python3
#!/usr/bin/env python3                                                                                                                                                                                 

import time                                                                                                                                                                                            
import qubesadmin                                                                                                                                                                                      

ROUND_PRECISION = 3                                                                                                                                                                                    

def get_time():                                                                                                                                                                                        
    return time.clock_gettime(time.CLOCK_MONOTONIC)                                                                                                                                                    

def call_api(service, qube):                                                                                                                                                                           
    start_time = get_time()                                                                                                                                                                            
    app = qubesadmin.Qubes()                                                                                                                                                                           
    domains = app.domains                                                                                                                                                                              
    appvm = domains[qube]
    domain_time = get_time()
    target_qube = qubesadmin.vm.DispVM.from_appvm(app, appvm)
    name = target_qube.name
    # A very small number, if it appears, it will show a bottleneck at
    # DispVM.from_appvm.
    target_time = get_time()
    try:
        target_qube.run_service_for_stdio(service, timeout=60)
    except subprocess.CalledProcessError as e:
        raise Exception(
            f"'{name}': service '{service}' failed ({e.returncode}):"
            f" {e.stdout},"
            f" {e.stderr}"
        )
    except subprocess.TimeoutExpired as e:
        raise Exception(
            f"'{name}': service '{service}' failed: timeout expired:"
            f" {e.stdout},"
            f" {e.stderr}"
        )
    run_service_time = get_time()
    target_qube.cleanup()
    cleanup_time = get_time()
    end_time = cleanup_time
    runtime = {}
    runtime["dom"] = round(domain_time - start_time, ROUND_PRECISION)
    runtime["disp"] = round(target_time - domain_time, ROUND_PRECISION)
    runtime["exec"] = round(run_service_time - target_time, ROUND_PRECISION)
    runtime["clean"] = round(
        cleanup_time - run_service_time, ROUND_PRECISION
    )
    runtime["total"] = round(end_time - start_time, ROUND_PRECISION)
    return runtime

def main():
    runtime = call_api("qubes.WaitForSession", "default-dvm-gui")
    print("GUI:", gui_runtime)

if __name__ == "__main__":
    main()
```

marmarek:
@ben-grande I added announcing `late-gui-daemon` feature, should be visible as `supported-feature.late-gui-daemon` feature on the template.

ben-grande:
> I added announcing `late-gui-daemon` feature, should be visible as `supported-feature.late-gui-daemon` feature on the template.

Using it now.


> The graph above shows `unbound-anchor.service`, any idea what is it waiting for? Or whether we need it at all?

It seems to be a timer triggered daily, this is not nice for disposables as it is triggered on every new one until it runs on the template or it is outside of the time it runs.

```
[user@disp6051 ~]$ systemctl status unbound-anchor
___ unbound-anchor.service - update of the root trust anchor for DNSSEC validation in unbound
     Loaded: loaded (/usr/lib/systemd/system/unbound-anchor.service; static)
    Drop-In: /usr/lib/systemd/system/service.d
             ______10-timeout-abort.conf
     Active: inactive (dead)
TriggeredBy: ___ unbound-anchor.timer
       Docs: man:unbound-anchor(8)
[user@disp6051 ~]$

[user@disp6051 ~]$ systemctl status 'unbound*'
___ unbound-anchor.timer - daily update of the root trust anchor for DNSSEC
     Loaded: loaded (/usr/lib/systemd/system/unbound-anchor.timer; enabled; preset: enabled)
     Active: active (waiting) since Sat 2025-11-29 18:01:45 UTC; 7min ago
 Invocation: 4e65589edde64399ae266f22dbedb19d
    Trigger: Sun 2025-11-30 00:00:00 UTC; 5h 50min left
   Triggers: ___ unbound-anchor.service
       Docs: man:unbound-anchor(8)
```

> > qubes-notification-agent shows `2.7` seconds

I am logged in on dom0, it still takes that time, some times more, like 15 seconds. But even when it takes 2 seconds, the `qubes.WaitForSession` is slow... unsure why.

> INFO: vm.disp9848: Preload startup log '(b'running\n111ms xfconfd.service\n 43ms at-spi-dbus-bus.service\n 33ms gvfs-daemon.service\n 19ms dbus-broker.service\n  9ms systemd-tmpfiles-setup.service\n  9ms dconf.service\n  6ms dbus.socket\n', b'\nreal\t0m45.120s\nuser\t0m0.003s\nsys\t0m0.004s\n')'

```
systemd-analyze --user
```

Constantly shows 45s to reach default.target. I could see that from the plot, but it doesn't show what is taking so long.

```
systemd-analyze --user critical-chain
```

Also doesn't show the culprit.

But when looking at the logs:

```
journalctl --user
```

It shows that qubes-notification-agent timed out the first time it tried to start, it took 45 seconds, then it attempts a restart and it works.

`man systemd.service` mentions `TimeoutStartSec` and `DefaultTimeoutStartSec`, which on `man systemd-system.conf`, shows the value of 45s.

The service definitely starts too early. So it needs atleast `After=qubes-qrexec-agent.service`, but that doesn't seem to be enough, there is something else missing... What helped was reducing `TimeoutStartSec` to `5s`, but it seems the service fails to start twice now, the `default.target` can be reached in `5s` now, it can be reduced even more, like `2s`.

Changing `TimeoutStartSec` or not, it fails on the first try, meaning that a very early request will fail:

```
user@dom0:~$ qvm-run -p --dispvm=default-dvm-gui 'notify-send hey there'
GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: Could not activate remote peer 'org.freedesktop.Notifications': activation request failed: unit is invalid
```

The service `qubes-notification-agent` will restart, but the command above doesn't wait for it to be ready and fails.

> Hm, if I read the code correctly, it waits for dom0 part to initialize, which can happen only when dom0 user is logged in... For preloaded, it's okay to not wait for this one, but I'm not sure how to do that in practice. Adjust service ordering to not have some Before=? or modify some of the `qubes.Wait*` services?

That is indeed not nice for preloaded disposables. Don't know yet what to do about that...

ben-grande:
:) Things are working.

- [dispvm](https://openqa.qubes-os.org/tests/161056): Some failures on tests that have a lot of disposables for virtualized Qubes to handle
- [dispvm_perf](https://openqa.qubes-os.org/tests/161078): 2 failures on Whonix, 1 was with a lot of preloaded disposables, the other was timeout trying to shutdown Whonix disposable template, which is done on `setUp`

PR #256, opened 2026-02-03
Remove duplicate buttons using xfconf-query

noskb:
This will remove duplicate close buttons and others from affected
applications such as gedit in Xfce-based Debian/Fedora templates.

At first, I tried adding it to `60xfce-desktop.sh`, but it had no effect.


appvm-scripts/etc/xdgautostart/qubes-set-xsettings-xfconf-query.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Version=1.0
+Encoding=UTF-8
+Name=xfconf-query for Xfce-based QubesVM
+Exec=/usr/bin/xfconf-query --channel xsettings --property /Gtk/DecorationLayout --set ""
```
ben-grande: This is assuming that the qube:

1. has `xfconf-query`
2. is currently using Xfce

Both conditions need to be met for this script to be called.

First, can be done with `TryExec=/usr/bin/xfconf-query`. Also, I'd rather the path was not hardcoded.

Second, `XDG_SESSION_DESKTOP=X-QUBES` is present on `fedora-43-minimal`, which is not the Xfce flavor and doesn't have `xfconf-query`. So you will need to find another thing to query.

ben-grande (changes requested):

appvm-scripts/etc/xdgautostart/qubes-set-xsettings-xfconf-query.desktop
noskb: Thank you for your review.

Since this won't take effect unless the `qubes-gui-agent-xfce` is installed, I think it's reasonable to assume that `xfconf-query` exists.  That’s why I didn’t use `TryExec=`, but I wonder if I’m missing something.
<br>
 > Also, I'd rather the path was not hardcoded.

agree.

<br>
Since minimal template is intended for advanced users, I hadn't considered it.

I expect the same thing is possible by adding  `xsettingsd` to the `qubes-gui-agent` package dependencies and placing the [configuration file](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/appvm-scripts/etc/xdg-debian/xsettingsd) in its proper location, but I’m hesitant to add anything by default to minimal template, which is for advanced users who prefer minimizing things…

ben-grande: > Since minimal template is intended for advanced users, I hadn't considered it.

For advanced users to configure, doesn't mean things should be broken. I don't know if this breaks systemd somehow if it is configured to handle xdg autostart. But I'd rather that it is not left in a failed state, but skipped.

> I expect the same thing is possible by adding `xsettingsd` to the `qubes-gui-agent` package dependencies and placing the [configuration file](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/appvm-scripts/etc/xdg-debian/xsettingsd) in its proper location, but I’m hesitant to add anything by default to minimal template, which is for advanced users who prefer minimizing things…

No need to add the package. It doesn't guarantee that xfce is currently running. Please find a way to know if Xfce is currently running and add that as a check to `TryExec=`.

ben-grande: > > Since minimal template is intended for advanced users, I hadn't considered it.
> 
> For advanced users to configure, doesn't mean things should be broken.

To explain this better, minimal templates don't have everything working by default, I just don't want a failed state by default.

noskb: >No need to add the package. It doesn't guarantee that xfce is currently running. Please find a way to know if Xfce is currently running and add that as a check to TryExec=.

Is it really needed to check whether Xfce is currently running here?

All environment variables recommended for checking whether XFCE (DE) is currently running are override  by [`60xfce-desktop`](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/appvm-scripts/etc/X11/Xsession.d/60xfce-desktop) included in the `qubes-gui-agent-xfce` package, so as I see it, if the package is installed, Xfce is assumed to be running. Furthermore,`xfconf-query` does not fail even if Xfce is not running. But, `--set` alone will fail if the channel or property does not exist, so adding `--create` is planned.`)

``` .desktop
TryExec=xfconf-query
Exec=xfconf-query --channel xsettings --property /Gtk/DecorationLayout --create --type string --set ""

````
With this change, it should no longer default to a failed state. (for example, even in a unique environment where only `xfconf` and `qubes-gui-agent-xfce` packages are installed on minimal template)

Comment (marmarek) [reply appvm-scripts/etc/xdgautostart/qubes-set-xsettings-xfconf-query.desktop:5]: `TryExec` approach is IMO good enough

**Comment by ben-grande** [comment]
Please remove the merge commit. I normally rebase against the QubesOS remote definition to avoid a merge commit.
```
% git remote -v
origin  gh:ben-grande/qubes-app-linux-usb-proxy (fetch)
origin  gh:ben-grande/qubes-app-linux-usb-proxy (push)
upstream        gh:qubesos/qubes-app-linux-usb-proxy (fetch)
upstream        gh:qubesos/qubes-app-linux-usb-proxy (push)

git rebase -i upstream/main
```

Creates no merge commit.

noskb:
I'm a bit confused because I'm not familiar withgit. I think creating a new PR will help avoid cluttering the commit history, so I'll go ahead and create a new one.

noskb:
Closing this because it has been replaced by https://github.com/QubesOS/qubes-gui-agent-linux/pull/261

PR #257, opened 2026-04-12
Sync environment variables from the user session into systemd and dbus

### LINKED ISSUE #10712: fedora-43 Terminal (ptyxis) crash
### Qubes OS release

`4.3.0`.

### Brief summary

I created a new qube, `work2`, based on a fresh `fedora43` template. I did install some packages on the template, also modify the scaling and dpi settings, but otherwise it's pretty much clean. No in-place update. 

The terminal is very flaky. When I start a terminal via the launcher (I use i3), and the terminal pops up, it will crash if I hit `shift-ctrl-t` (new tab) a few times, or on `shift-ctrl-n` (new window). 

However, if I start `xterm`, and via `xterm` launch the terminal via `ptyxis`, I can _not_ make it crash.

### LINKED ISSUE #10759: systemd's user manager's environment does not match up with the user session's environment, causing incorrect behavior in some situations
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### Qubes OS release

R4.3

### Brief summary

This is basically the sequel to https://github.com/QubesOS/qubes-issues/issues/10299.

When fixing the above bug report, we successfully made it so that the environment variables systemd's user manager sees are sync'd into the user session, without clobbering anything. However, there is not yet anything that syncs the user session's variables back into the systemd user manager. This means that anything that depends on systemd's user manager's environment variables may behave differently than the user expects.

Case in point, I recently wrote a shim for the `org.freedesktop.FileManager1` D-Bus interface, to work around [this mess](https://forums.whonix.org/t/is-catfish-the-new-default-in-the-qubes-domain-open-file-manager-quick-widget-for-whonix/20233/22). It consists of a backend that listens for D-Bus method calls, and a frontend that allows the user to permit or deny a folder open operation and that launches the default file manager when requested. In order to ensure that the shim  starts up before the qube launches any applications requested by the user, it is currently installed as a user service in systemd. This means it inherits systemd's user manager's environment.<sup>1</sup> When a method call is received, it launches the frontend, and if the user permits the folder open operation, the default file manager is opened...

...except it's not. systemd's user manager's `XDG_DATA_DIRS` doesn't include `/usr/share/usability-misc/xdg-override/` in it, and so the config file that sets the default file manager to PCManFM-Qt is missed. As a result, the default file manager appears to be Catfish instead, and that's what ultimately gets launched, which is not at all what the user expects.

<sup>1: Well, kinda anyway. Because [graphical-session.target doesn't work](https://github.com/QubesOS/qubes-issues/issues/9576), it is possible the shim backend will start before X.org starts, so in reality the shim has code in it that fetches the systemd user manager's current environment before launching the shim frontend, and then feeds that environment into the frontend.</sup>

### Steps to reproduce

1. Launch QTerminal in a Kicksecure 18 AppVM.
2. Run `systemctl --user show-environment | grep -i XDG_DATA_DIRS`.

### Expected behavior

`XDG_DATA_DIRS` should contain `/usr/share/usability-misc/xdg-override/`.

### Actual behavior

`XDG_DATA_DIRS` does not contain `/usr/share/usability-misc/xdg-override/`.

### Additional information

There is a utility for fixing this exact issue, `dbus-update-activation-environment` (specifically with the `--systemd` and `--all` options), which I believe is preinstalled on many if not all official VM templates for Qubes. If this was run by something in the user session during startup, it would likely fix this issue. I originally suggested [something like this](https://github.com/QubesOS/qubes-gui-agent-linux/commit/e4c68aaf51a64c8b33e47123f42f1669001fe752) to fix the earlier environment variable bug linked above, but [it was rejected](https://github.com/QubesOS/qubes-gui-agent-linux/pull/246#issuecomment-3369522903) due to the fact that it would clobber environment variables from `/etc/environment.d`. Now that variables are synced from systemd into the user session, maybe we can get away with using `dbus-update-activation-environment --systemd --all` at some point *after* the variables are synced by `qubes-gui-runuser`? We're syncing them in one direction already, I think we just need to sync them the other way too.

This *might* be related to https://github.com/QubesOS/qubes-issues/issues/10644, although that may be far-fetched since I don't *think* systemd's user manager is in charge of launching applications when using debian-13-xfce...

ArrayBolt3:
We're already syncing environment variables from the systemd user manager into the user session, but some variables may be set in the user session that aren't in the user manager or dbus-daemon by default. This interferes with applications that are launched by the user manager or dbus-daemon, such as Ptyxis and Kicksecure's fm-shim system.

Sync environment variables from the user session into the systemd user manager and dbus-daemon to fix this.

Fixes: https://github.com/QubesOS/qubes-issues/issues/10759
Fixes: https://github.com/QubesOS/qubes-issues/issues/10712

(Note that I haven't finished testing if this fixes the Ptyxis terminal issue, but it seems like it should. I'll report back once I've tested that.)

### DIFF
--- a/appvm-scripts/usrbin/qubes-session
+++ b/appvm-scripts/usrbin/qubes-session
@@ -51,6 +51,10 @@ VMTYPE=$(/usr/bin/qubesdb-read /qubes-vm-type)
 UPDTYPE=$(/usr/bin/qubesdb-read /qubes-vm-updateable)
 [[ $UPDTYPE == 'True' ]] && UPDTYPE="UpdateableVM" || UPDTYPE="NonUpdateableVM"
 
+# Sync environment variables from the user session into the systemd user
+# manager and dbus-daemon
+dbus-update-activation-environment --systemd --all
+
 # process /etc/xdg/autostart and friends (according to Desktop Application
 # Autostart Specification)
 /usr/bin/qubes-session-autostart QUBES X-QUBES "X-$VMTYPE" "X-$UPDTYPE"


ben-grande:
Can some test be added that covers the initial issue and the current issue:

- don't clobber env vars
- but do set them...

I think it will have to be an integration test.

ArrayBolt3:
@ben-grande Where are the integration tests stored?

ben-grande:
I never did an integration test outside of core-admin, but here is an example of pdf sanitizer using `qubes.tests.extra`:

https://github.com/QubesOS/qubes-app-linux-pdf-converter/blob/main/qubespdfconverter/tests/__init__.py

ben-grande:
Did not find an error on openqa related to this PR.

ArrayBolt3:
Created https://github.com/QubesOS/qubes-core-admin/pull/804 for the integration tests, will probably need partially re-worked before it can be merged.

marmarek (approved):

PR #258, opened 2026-04-22
qubes-keymap.sh: preserve user layout across NetVM changes

nancysangani:
Fixes QubesOS/qubes-issues#9892

When NetVM is changed, gui-daemon re-writes /keyboard-layout in QubesDB with the GuiVM default. The watch loop in qubes-keymap.sh fires and blindly applies it, resetting any custom layout set via setxkbmap.

Fix by reading the full live X11 layout (layout+variant+options) via setxkbmap -query on watch events, so user customizations survive NetVM changes. Falls back to QubesDB only when no custom layout is active. Also honours ~/.config/qubes-keyboard-layout.rc written by qubes-change-keyboard-layout.

### DIFF
--- a/appvm-scripts/usr/lib/qubes/qubes-keymap.sh
+++ b/appvm-scripts/usr/lib/qubes/qubes-keymap.sh
@@ -2,6 +2,9 @@
 
 # This file may be also executed by qubes-change-keyboard-layout
 
+# User-saved layout file (written by qubes-change-keyboard-layout)
+CUSTOM_LAYOUT_FILE="${HOME:-/root}/.config/qubes-keyboard-layout.rc"
+
 set_keyboard_layout() {
     KEYMAP="$1"
     # Default value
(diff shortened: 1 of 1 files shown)


ben-grande:
Please make separate commits for the fix and the cleanup.

nancysangani:
> Please make separate commits for the fix and the cleanup.

Fixed!

appvm-scripts/usr/lib/qubes/qubes-keymap.sh
```diff
@@ -15,32 +17,60 @@ set_keyboard_layout() {
...
     fi
-
     if [ -n "$KEYMAP_VARIANT" ]; then
         KEYMAP_VARIANT="-variant $KEYMAP_VARIANT"
     fi
-
     if [ -n "$KEYMAP_OPTIONS" ]; then
         KEYMAP_OPTIONS="-option -option $KEYMAP_OPTIONS"
     fi
+    for x in /tmp/.X11-unix/X*; do
```
ben-grande, line 26: This directory doesn't exist on domUs.

ben-grande (changes requested):

appvm-scripts/usr/lib/qubes/qubes-keymap.sh
marmarek: It does, it's a standard X11 socket location

ben-grande: I missed the `.`...

nancysangani: @marmarek thanks for checking - glad it's confirmed correct.

ben-grande:
> When NetVM is changed, gui-daemon re-writes /keyboard-layout in QubesDB with the GuiVM default.

I don't think the qubesdb for keyboard should be reset when changing netvm. I think the fix should be somewhere else. Although I cannot find an issue in `qubes.ext.gui` just from looking at the code, would need to add some debugging.


Edit: domain-qdb-create is fired.

nancysangani:
@ben-grande Thanks for looking into this. So the root cause is that `domain-qdb-create` is fired on NetVM changes, which resets `/keyboard-layout`. Should the fix be in `qubes-gui-daemon` to skip re-writing `/keyboard-layout` if it already has a value, or in `qubes-core-admin` to not fire `domain-qdb-create` for keyboard-related keys on NetVM changes? Happy to move the fix there once the right location is confirmed.

ben-grande:
Do have QubesOS running to test things?

marmarek:
> Should the fix be in `qubes-gui-daemon` to skip re-writing `/keyboard-layout` if it already has a value, or in `qubes-core-admin` to not fire `domain-qdb-create` for keyboard-related keys on NetVM changes?

I'd say the latter. It sounds like it should be limited to update only network-related entries, not all of them.

nancysangani:
> Do have QubesOS running to test things?

@ben-grande No, I don't have Qubes OS running currently. I was working from code reading alone.

nancysangani:
> > Should the fix be in `qubes-gui-daemon` to skip re-writing `/keyboard-layout` if it already has a value, or in `qubes-core-admin` to not fire `domain-qdb-create` for keyboard-related keys on NetVM changes?
> 
> I'd say the latter. It sounds like it should be limited to update only network-related entries, not all of them.

Thanks for the direction. I'll close this PR and open a new one in `qubes-core-admin`. One question before I do — should ``on_domain_qdb_create skip all non-network entries when triggered by a NetVM change, or is skipping `/keyboard-layout` specifically sufficient for now?

marmarek:
I'd say there shouldn't be `create_qdb_entries()` call on netvm change, but instead it should call a function that does only network-related things

nancysangani:
@marmarek Understood — so the fix is to extract the network-related QDB writes into a separate function and call that on NetVM changes, instead of calling the full `domain-qdb-create`. I'll look into where NetVM changes trigger `domain-qdb-create` and implement that. Opening the new PR in `qubes-core-admin`.

PR #259, opened 2026-05-03
Fix cursors never changing in Chromium apps and derivatives

### LINKED ISSUE #7238: chromium/electron-based apps can't change cursor shape
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### Qubes OS release

r4.1

### Brief summary

Cursor change support is added in #1551. However, it doesn't work with electron-based apps. Other applications, like GTK-based or Firefox work as intended with correct mouse cursor shape on hovering text and buttons.

### Steps to reproduce

1. install arch linux in AppVM
2. install `electron`, `code` with `pacman`

### Expected behavior

The mouse cursor should change when hovered over text or buttons.

### Actual behavior

The mouse cursor is always the same inside electron-based apps.

t-grm:
Main issue thread: https://github.com/QubesOS/qubes-issues/issues/7238
More info on the fix: https://github.com/QubesOS/qubes-issues/issues/7238#issuecomment-4364826060

A useful test page for before/after comparison: https://vibhorjaiswal.github.io/Cursor-Test/

Chromium, unfortunately, never calls XfixesSetCursorName() from their side, so while a XfixesDisplayCursorNotify event is eventually received by the qubes-gui-agent, cursor_name is always empty and the agent always falls back to a default cursor.

This PR fixes this behaviour by patching the qubes-gui-agent service to be a little bit smarter: whenever we receive an XfixesDisplayCursorNotify event with no cursor_name attached, instead of immediately returning the default cursor we'll:

- fetch the actual bitmap image of the new cursor with XFixesGetCursorImage(g->display)
- build a hash based on the bitmap's dimensions and shape
- loop over the list of supported cursors, hash each one of them in the same way and try to match them against our candidate
- return the matched cursor or fall back to the default cursor

It is not ideal in that at every cursor swap without a specified cursor_name we now need to do additional work (notably load NUM_SUPPORTED_CURSORS bitmaps), but the added computation time should be negligible in most cases as cursor changes should be relatively infrequent in the grand scheme of things, their bitmaps should be tiny, the hash function is quick, we're probably using the page cache after the first hit, etc.

Ideally, Chromium would systematically pass the cursor's name properly...

TODO: I initially wanted to precompute the hashes of all supported cursors in the supported_cursors variable to avoid redundant computations, but it didn't work because of differing cursor dimensions between the time we precompute and the time we check.

Disclaimer: I used Claude chat to recommend a quick & simple hashing function (https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) and then to review the patch. In particular it confirmed a potential issue with unbounded checks on the cursor's size, fixed in the second commit.


marmarek:
> TODO: I initially wanted to precompute the hashes of all supported cursors in the supported_cursors variable to avoid redundant computations, but it didn't work because of differing cursor dimensions between the time we precompute and the time we check.

If that doesn't work, then maybe cache last computed hash list, together with the size (and theme?) - and next time compare size+theme to see if cache can be reused or needs rebuild.

I would assume in most (all?) cases Chromium would use the same size/theme for every cursor, so it should be possible  to calculate the hashes once.

marmarek:
See also CI output - it can't find X11/Xcursor/Xcursor.h - missing build dependency?

t-grm:
Ah yes, that's the added #include. I manually installed `libxcursor-dev` on my Debian VM, which is `libXcursor-devel` on Fedora.

Should I simply add `BuildRequires: libXcursor-devel` in `/qubes-gui-agent-linux/rpm_spec/gui-agent.spec.in` and push the change to this PR for the CI pipeline to refresh?

marmarek:
> Should I simply add `BuildRequires: libXcursor-devel` in `/qubes-gui-agent-linux/rpm_spec/gui-agent.spec.in` and push the change to this PR for the CI pipeline to refresh?

Yes. And similarly in debian/control.

marmarek:
BTW, I prefer amending commit with such fix + force-push

t-grm:
Hi @marmarek, the build is looking good! There's a failure in one of the repro tasks but it's marked as "allowed to fail" for some reason.

I've added a precomputation pass that only runs the first time we hit an unnamed cursor, so well-behaving apps won't ever see it.

I also ran a few basic profiling tests: `precompute_hashed_cursors` takes about 9ms once, the first time we swap an unnamed cursor; then `find_cursor_by_image` is only about 40μs per subsequent _unnamed_ cursor change.

marmarek (approved):

PR #260, opened 2026-05-23
Fix registering user session in sys-gui

### LINKED ISSUE #10890: Non-dom0 Interface qubes don't have lockable sessions with `loginctl lock-sessions`
### Qubes OS release

Qubes OS 4.3

### Brief summary

As per the conversation at https://github.com/BusKill/buskill-qubes-contrib/pull/1#discussion_r3249530543, running `loginctl lock-sessions` in interface qubes other than dom0 (confirmed in sys-gui and sys-gui-vnc) is a no-op. None of the sessions listed with `loginctl list-sessions` can be locked, except running `xflock4`, assuming the interface qubes are based on the fedora-42-xfce TemplateVM.

### Steps to reproduce

1. Install Qubes OS from the Qubes-R4.3.0-x86_64.iso image and have initial-setup create Fedora 42 (fedora-42-xfce) TemplateVM.
2. Create an interface qube as described [in the documentation](https://doc.qubes-os.org/en/r4.3/user/advanced-topics/gui-domain.html), sys-gui in this example
3. In lightdm log in by setting the session to "GUI Domain (sys-gui)"
4. When the interface qube has finished loading, open a terminal in the interface qube and run `loginctl lock-sessions`.

### Expected behavior

The interface qube has the graphical session locked, just like when running `xflock4`.

### Actual behavior

The interface qube doesn't lock the graphical session:

```shell
[user@sys-gui ~]$ loginctl list-sessions 
SESSION  UID USER SEAT LEADER CLASS      TTY IDLE SINCE
      1 1000 user -    698    background -   no   -    
      2 1000 user -    702    manager    -   no   -    
      4 1000 user -    868    background -   no   -    

3 sessions listed.
[user@sys-gui ~]$ loginctl lock-session 1
Failed to issue method call: Session does not support lock screen.
[user@sys-gui ~]$ loginctl lock-session 2
Failed to issue method call: Session does not support lock screen.
[user@sys-gui ~]$ loginctl lock-session 4
Failed to issue method call: Session does not support lock screen.
```

### Additional information

Both dom0 and sys-gui do have the `xss-lock xflock4` process running.

For context, this issue has been observed when working on resolving https://github.com/QubesOS/qubes-issues/issues/10567, one of the requirements being described at https://opencollective.com/buskill/projects/qubes-package:

> Installing the BusKill contrib package should require the user to execute just one command after a fresh install of the latest stable version of QubesOS

This has been observed after installing Qubes OS from the Qubes-R4.3.0-x86_64.iso image and not applying any updates, as per the requirement. If this issue is no longer applicable with the most recent updates applied, this ticket can be closed.

marmarek:
Starting Xorg for sys-gui (on display :1) intentionally skips
qubes-gui-runuser and uses plain runuser - which doesn't register
graphical session in systemd-logind. This is okay, because actual user
session is running under Xephyr on display :0. But starting Xephyr was
missing qubes-gui-runuser call too.

Not registered user session caused at least `loginctl lock-session` not
working. There might be more issues caused by this.

Fix this by adding qubes-gui-runuser in qubes-start-xephyr script. This
is a bit more involved, because:
1. When started as qrexec service (which is the case normally), it's
   already running under registered user session, and logind refuses to
   create another one for that process. Fix this by using systemd-run,
   to run the actual service under new systemd unit, without user
   session yet, on a fresh terminal (tty8, as tty7 is used by Xorg :1
   already).
2. Non-privileged user is not allowed to create new sessions. Fix this
   by adding force-user='root' to the service configuration (and
   removing check that refuses root).

Finally, the qubes-gui-runuser call is a bit complicated, because
qubes-gui-runuser needs to have DISPLAY=:0 - to register this one in
logind. But then, Xephyr needs to be started with DISPLAY=:1 to display
its window there. Add extra 'env' call in between to adjust DISPLAY
variable appropriately.

Fixes https://github.com/QubesOS/qubes-issues/issues/10890

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,8 @@ endif
 		$(DESTDIR)/etc/qubes-rpc/qubes.SetMonitorLayout
 	ln -sf ../../usr/bin/qubes-start-xephyr \
 		$(DESTDIR)/etc/qubes-rpc/qubes.GuiVMSession
+	install -d $(DESTDIR)/etc/qubes/rpc-config
+	echo "force-user='root'" > $(DESTDIR)/etc/qubes/rpc-config/qubes.GuiVMSession
 	install -D window-icon-updater/icon-sender \
 		$(DESTDIR)/usr/lib/qubes/icon-sender
 	install -m 0644 -D window-icon-updater/qubes-icon-sender.desktop \
--- a/appvm-scripts/usrbin/qubes-start-xephyr
+++ b/appvm-scripts/usrbin/qubes-start-xephyr
@@ -1,8 +1,15 @@
 #!/bin/sh
 
-if [ $(whoami) == "root" ]; then
-    echo "Should be run as normal user!"
-    exit 1
+# Start under new service unit, to detach from the current non-gui logind
+# session
+if [ -z "$INVOCATION_ID" ]; then
+    exec systemd-run --system \
+        --unit=qubes-gui-agent-session.service \
+        --description="Start GUI session via Xephyr" \
+        --property=TTYPath=/dev/tty8 \
+        --property=StandardInput=tty \
+        --wait \
+        -- "$0" "$@"
 fi
 
 # do not let child processes believe they are running as qrexec service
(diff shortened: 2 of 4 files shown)


marmarek:
openQArun MACHINE=64bit,hw13 TEST=system_tests_gui_interactive,system_tests_guivm_gpu_gui_interactive,system_tests_guivm_gui_interactive,system_tests_guivm_vnc_gui_interactive,system_tests_i3_gui_interactive,system_tests_kde_gui_interactive

PR #261, opened 2026-06-07
Remove duplicate buttons using xfconf-query for xfce-based VM

noskb:
In the gnome-based template, xsettingsd was used to remove duplicate buttons. However, since xfce-based templates use xfsettingsd, xfconf-query is called at session startup to perform the same configuration.

Previously: https://github.com/QubesOS/qubes-gui-agent-linux/pull/256

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,6 +239,8 @@ endif
 		$(DESTDIR)/usr/lib/qubes/qubes-xorg-wrapper
 	install -D appvm-scripts/usr/lib/qubes/gtk4-workarounds.sh \
 		$(DESTDIR)/usr/lib/qubes/gtk4-workarounds.sh
+	install -m 0644 -D appvm-scripts/etc/xdgautostart/qubes-set-xsettings-xfconf-query.desktop \
+		$(DESTDIR)/etc/xdg/autostart/qubes-set-xsettings-xfconf-query.desktop
 ifeq ($(shell lsb_release -is), Debian)
 	install -D -m 0644 appvm-scripts/etc/pam.d/qubes-gui-agent.debian \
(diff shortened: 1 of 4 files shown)


marmarek:
Hm, looking at this more, there is a significant issue with this approach: it also disables buttons in ~GUI domain~ Interface qube, for applications running directly there (like xfce settings windows). This is because the setting set via `xfconf-query` will persist in user home, so even if at later time user session is started with `XDG_CURRENT_DESKTOP=XFCE` (so `OnlyShowIn=X-QUBES;` will prevent calling `xfconf-query`), the setting will be carried over from previous startup.

So, I'm afraid it needs to be a bit smarter. Maybe a script that saves old value to a file in user home (somewhere in `~/.config`)  and on non-qubes session restores it?

noskb:
> the setting set via xfconf-query will persist in user home

This is a strange behavior of XFCE, but when only the `/Gtk/DecorationLayout` is changed, it is not persisted. It appears that the value reverts to`/general/button_layout` value of `xfwm4` property (even if it does not exist, revert to the default) when xfsettingsd starts up. And that was exactly why the changes didn't take effect even after running xfconf-query in [60xfce-desktop](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/appvm-scripts/etc/X11/Xsession.d/60xfce-desktop), which is why I tried using /etc/xdg/autostart as an alternative...


> So, I'm afraid it needs to be a bit smarter. Maybe a script that saves old value to a file in user home (somewhere in ~/.config) and on non-qubes session restores it?

Wouldn't it be sufficient to simply skip it when guivm service is enabled?

PR #264, opened 2026-07-20
Fix XEmbed tray docking initialization

besendorf:
## Summary

Fix XEmbed tray clients that remain mapped as blank `1x1` icons when proxied to a desktop tray.

The live tray-docking path now:

- sends the constructed `XEMBED_EMBEDDED_NOTIFY` event instead of the address of the incoming-event pointer;
- reads `XEMBED_MAPPED` using the `unsigned long` representation returned by `XGetWindowProperty()`;
- sizes the client to the existing `32x32` default embedder geometry; and
- sends that geometry to the GUI daemon before `MSG_DOCK`, matching the ordering used when full window state is resent.

## Root cause

The incorrect `XSendEvent()` argument prevented clients from receiving `XEMBED_EMBEDDED_NOTIFY`.

Fixing only that call was insufficient for clients initially created at `1x1`: dom0 still knew the pre-dock window as `1x1` and configured the docked client back to that size. Additionally, reading a format-32 X property through `int *` selects the wrong slot on 64-bit systems, so the mapped flag could be missed.

## Validation

- Rebuilt `gui-agent/qubes-gui` from a clean object state with the project's `-Wall -Wextra -Werror` flags.
- Tested on Qubes OS 4.3 with Fedora 43 guests and KDE Plasma/X11 plus `xembedsniproxy` in dom0.
- A minimal XEmbed client initially created at `1x1` changed from receiving no embedded notification and remaining `1x1` to receiving `XEMBED_EMBEDDED_NOTIFY`, `MapNotify`, and a stable `32x32` configure event.
- Verified directly in the guest as `32x32` and `IsViewable`.
- Tested the resulting agent in `sys-net` with the real `nm-applet`; the previously blank NetworkManager icon became visible and was tinted with the qube's red label color.

## AI assistance disclosure

This change was developed and tested with assistance from OpenAI Codex. The runtime tests and final visual confirmation were performed interactively on the affected Qubes OS system.

### DIFF
--- a/gui-agent/vmside.c
+++ b/gui-agent/vmside.c
@@ -1410,6 +1410,7 @@ static void process_xevent_message(Ghandles * g, XClientMessageEvent * ev)
         Window w;
         int ret;
         struct msg_hdr hdr;
+        struct msg_configure conf;
         Atom act_type;
         int act_fmt;
         int mapwindow = 0;
@@ -1442,7 +1443,7 @@ static void process_xevent_message(Ghandles * g, XClientMessageEvent * ev)
                     fprintf(stderr, "window 0x%lx havn't proper _XEMBED_INFO property, assuming defaults (workaround for buggy applications)\n", w);
                 }
                 if (act_type == g->xembed_info && nitems == 2) {
-                    mapwindow = ((int*)data)[1] & XEMBED_MAPPED;
+                    mapwindow = ((unsigned long*)data)[1] & XEMBED_MAPPED;
                     /* TODO: handle version */
                 }
                 if (ret == Success && nitems > 0)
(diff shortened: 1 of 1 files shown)


gui-agent/vmside.c
```diff
@@ -1486,18 +1487,27 @@ static void process_xevent_message(Ghandles * g, XClientMessageEvent * ev)
...
                 XSync(g->display, False);
 
+                hdr.type = MSG_CONFIGURE;
+                hdr.window = w;
+                conf.x = 0;
+                conf.y = 0;
+                conf.width = 32;
+                conf.height = 32;
+                conf.override_redirect = 0;
+                write_message(g->vchan, hdr, conf);
```
marmarek, line 1509: This looks a bit risky, because the above `XMoveResizeWindow()` call will eventually also result in a `MSG_CONFIGURE` message (in response to X11 event). But, IIUC the order or messages is important here, and the other `MSG_CONFIGURE` will be send after `MSG_DOCK`. So, maybe duplicate message is less bad...

marmarek:
Finally the nm-applet icon is visible: https://openqa.qubes-os.org/tests/189082#step/clipboard_and_web/3 :)

marmarek:
Is it draft because of unverified commit signature? Or something else?

besendorf:
Converted. Can sign in about two weeks

besendorf:
@marmarek I signed the commits. Is anything else blocking this?

marmarek (approved):

ben-grande:
Some reports on the forum that running certain software when this patch is present, does slow down the system. More information on:

- https://forum.qubes-os.org/t/qubes-became-extremely-sluggish-after-a-dom0-update/42922/7

besendorf:
I will try to reproduce it. I have updated qubes-gui-agent in my appvms but not dom0 yet. I can see the icons and dont experience the performance issues mentioned in the thread so far

ben-grande:
Could you please try the mentioned program, `fcitx5` in an archlinux based qube? It seems that it is happening if those two conditions are met.

besendorf:
I could not reproduce with kde in dom0. Trying xfce next. Users in the forum mentioned high cpu usage for xfwm4

besendorf:
This fixes it for me: https://github.com/QubesOS/qubes-gui-agent-linux/pull/265

